PR #8, opened 2021-09-16
feat: created an rmarkdown doocuments which analyzes the initial ques…

charlie-costanzo:
…tions posed in issue #329


charlie-costanzo:
@machow -- regarding the PR above, the python code executes fine in rstudio but doesn't produce any output when knit together. would love to take a look at it with you when you get a chance

PR #25, opened 2021-11-18
More shared utils

tiffanychu90:
Renaming to `_shared_utils` to live at top of the repo.
Work in `example_reports` to resemble how analysts would use this package, which is in a sub-folder in `data-analyses`

GH issue: https://github.com/cal-itp/data-infra/issues/703


tiffanychu90:
@machow: can you look at this [notebook](https://github.com/cal-itp/data-analyses/blob/more-shared-utils/bus_service_increase/shared-utils-test.ipynb)? Right now, I run the `python setup.py install` at the root of the repo. Where did you mean to put the `-e .`? I tried `requirements.txt`, but that threw an error.

PR #36, opened 2021-12-13
Fix nbviewer workflow

thekaveman:
See jitterbit/get-changed-files#11 for context. 

Here's a failed run for a PR: https://github.com/cal-itp/data-analyses/runs/4371698104?check_suite_focus=true

When commits are pushed, then later squashed and re-pushed, the action can fail. Workaround is to continue on error, since the list of files changed is still correct.


thekaveman:
@edasmalchi @tiffanychu90 can someone please review and merge? I don't have write access to this repo.

PR #41, opened 2021-12-16
complete and summarize accessibility metrics

edasmalchi:


msd_dashboard_metric/01_area_population_metrics.ipynb
```diff
@@ -1481,9 +1085,137 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 59,
+   "execution_count": 45,
    "id": "14106a8c-c8d2-4a39-a372-111771905c64",
    "metadata": {},
+   "outputs": [],
+   "source": [
+    "rt_complete = (rt_complete >> select(_.calitp_itp_id == _.itp_id, _.calitp_url_number == _.url_number))"
```
evansiroky, line 1093: Is this needed? Seems like it's just selecting everything from itself.

edasmalchi: It doesn't select everything (dropping the various url columns for example). It also renames the itp_id and url_number columns to match the stops dfs for the join in the next cell

PR #45, opened 2021-12-21
Create script for data cleaning 

natam1:
- write script for data cleaning
- convert agency names using crosswalks 
- minor changes to cleaning notebooks


natam1:
@tiffanychu90 Here's the PR with my first script.

tiffanychu90:
@natam1: Functions are looking good within `clean_data.py`! 

Some additional things you might want to add in your next round of commits:
* You've defined the this function once above, and you can use it nested within another function: [code](https://github.com/cal-itp/data-analyses/blob/64982b49936c7ca267d1065374967aee1e010c89/dla/e76_obligated_funds/clean_data.py#L74-L81)
* Can shorten the prefix cleaning function a bit more with a couple of methods: [code](https://github.com/cal-itp/data-analyses/blob/64982b49936c7ca267d1065374967aee1e010c89/dla/e76_obligated_funds/clean_data.py#L99-L162) 

1. If you want to remove all the hyphens, using a `str.replace` should do the trick: 
`df["prefix"] = df.prefix.str.replace('-', '')`

1. For some of the bigger categories of replacements...do something like this: 
```
replace_me = ["CASB003", "CASB05", 'CASB06'] # fill out the rest of this list
for i in replace_me:
    df.prefix.replace(i, 'CASB', inplace=True)
    
replace_me = ['FERPL16', 'FERPL17'] # fill out the rest of this list
for i in replace_me:
    df.prefix.replace(i, 'FERPL', inplace=True)
```

Alternatively (this might be overkill):
```
# Dictionary method 
PREFIX_DICT_FIXES = {
    'CASB': ['CASB003', 'CASB05', 'CASB06'],
    'FERPL': ['FERPL16', 'FERPL17'],
}

for key, value in PREFIX_DICT_FIXES.items():
    # value is a list, so let's loop through each item in the list
    for i in value:
        # if it finds CASB003 (item within the list), replace it with CASB (key)
        df["prefix"] = df.apply(lambda x: key if x.prefix==i else x.prefix, axis=1)
```

PR #48, opened 2021-12-28
Stops and routes shapefiles

tiffanychu90:
* Create 2 geospatial datasets (zipped shapefiles? geojson?) to back a public-facing portal for Caltrans Traffic Operations. Goal is to see which traffic signals provide signal priority/signal preemption for buses
* Provide point geometry for all stops (merge in `route_id`, `route_name`, `agency_id`, `agency_name`)
* Provide line geometry for all routes (merge in to `route_id`, `route_name`, `agency_id`, `agency_name`)
* Do NOT export as ArcGIS Online hosted feature layer, which uses credits
* Set up as script to run in Airflow eventually, at some frequency (monthly?):  `export_shapefiles.py` is the one to be scheduled (note: it does depend on `shared_utils`)

To do later: Write function to overwrite an ESRI layer....will need credentials set up to that ArcGIS Server.

GH issue: https://github.com/cal-itp/data-infra/issues/889


tiffanychu90:
@charlie-costanzo: Do you have any ideas about how to make this query a bit more efficient? 

I did a `collect()` because I needed to do the `isin()`. But, later on, querying from the `gtfs_schedule.stop_times` table crashes the notebook/kubernetes(?), even though there's 2 inner joins that would massively pare down the queried result. I have `os.environ["CALITP_BQ_MAX_BYTES"] = str(100_000_000_000)`, and I'm not getting a max_bytes error anyway.

So far, I have it looping over the individual operators and appending. But, it's really only about <10 operators total, so I thought the inner join would have done the trick?
 
```
# Find the stops that aren't in `shapes.txt`
missing_trips = (
    tbl.gtfs_schedule.trips()
    >> select(_.calitp_itp_id, _.route_id, _.shape_id, _.trip_id)
    >> distinct()
    >> collect()
    >> filter(~_.shape_id.isin(routes_part1.shape_id))
)

def grab_missing_stops(ITP_ID, missing_trips_df, stops_df):
  df = (
      tbl.gtfs_schedule.stop_times()
      # Loop through individual operators, then do the join to find those trips
      # until the issue that makes kernel restarts is fixed
      >> filter(_.calitp_itp_id == ITP_ID)
      >> select(_.calitp_itp_id, _.stop_id, _.stop_sequence, _.trip_id)
      # Can't do isin without the collect()
      # But collect() is what is making kernel restart / shutting down notebook
      >> distinct()
      >> collect()
      >> inner_join(_, missing_trips_df, ["calitp_itp_id", "trip_id"])
      >> inner_join(_, stops_df, ["calitp_itp_id", "stop_id"])
  )
  
  return df
```

Also, what's the difference between `list(missing_trips.calitp_itp_id.unique())` and `missing_trips.calitp_itp_id.unique().tolist()`?
`list(missing_trips.calitp_itp_id.unique())` doesn't work with the loop.  `missing_trips.calitp_itp_id.unique().tolist()` does work with the loop.

Notebook here: https://github.com/cal-itp/data-analyses/blob/shapefiles/traffic_ops/stops-routes-shapefiles.ipynb

charlie-costanzo:
@tiffanychu90 thanks for tagging me in this! Let me take a look and see what I can find. I remember Michael saying that the need to `collect` so frequently in order to `isin` might be a bug in Siuba, so let me follow up with him on that

machow:
One quick thing I'm noticing (which doesn't seem like it'd be causing the crashing), joins need to use both calitp_itp_id and calitp_url_number, since they uniquely identify a feed (in hindsight I wish we'd made this a single column :/)

tiffanychu90:
> 

@machow: Thanks for the tip! In the future, will use both columns to join

PR #54, opened 2022-01-19
[do not merge] Functions for DLA

tiffanychu90:


tiffanychu90:
@edasmalchi: this [notebook](https://github.com/cal-itp/data-analyses/blob/functions/dla_functions/widgets.ipynb) has just one interactive dropdown widget, with a screenshot of where it updates in the logs, but not in the notebook

edasmalchi:
@tiffanychu90 Just pushed a commit that seems to fix, trick was calling the update function with the `with output:` context (see https://ipywidgets.readthedocs.io/en/latest/examples/Output%20Widget.html)

```
    def on_selection(*args):
        output.clear_output()
        with output:
            summarize_and_plot(df, select_col, dropdown.value)
```

tiffanychu90:
@edasmalchi: OOOH, thank you!

PR #63, opened 2022-02-01
Bus service workflow

tiffanychu90:
Reimagine the workflow for our analyses. 

Once an analysis has been presented and exploratory work is finished, the data creation and visualization steps should be moved from notebooks to scripts and be parameterized functions. Currently, it's hard for someone else to come in and reproduce visualizations. In theory, anyone should be able to come in and rerun those steps or subset the data, and produce the same set of visualizations, since one set of visualizations has already been done.

Should follow: 
1. Data creation (scripts only)
* Warehouse queries that are memory-intensive should sit separately. Stash locally or in GCS as parquets.
* Other `create_*.py` files that ingest, clean external data sources.
* Create analysis dataset with a `create_analysis_data.py` part 1 - assemble across disparate sources, save in GCS.
* `setup_viz_data.py` is part 2: further wrangle the analysis dataset to fit visualization needs, generate new metrics, etc etc 

2. Subset data
* Variety of ways to subset. Some are straightforward, use `isin` on some column that exists (district, county, certain operators). Others can be more complicated and require bringing in another geospatial dataset, doing a spatial join, etc.
* This step isn't fleshed out fully, and leave it as an explicit step to be done in a notebook or elsewhere. At any rate, the subset must have all its metrics generated in `setup_viz_data.py` already, it can only be pared down in this step.

3. Visualization
* Use `setup_*.py` to offload all the chart, map, interactive widget functions needed for visualizations.
* Use `make_*_viz.py` to compile the entire set of visualizations associated with a dataset or its subset. Ex: All the charts for CA operators are the same charts for D7 operators.

GH issue: https://github.com/cal-itp/data-infra/issues/1003


tiffanychu90:
@edasmalchi: Going through your notebooks to offload steps to fit the above defined workflow:

* [warehouse queries](https://github.com/cal-itp/data-analyses/blob/bus-service-workflow/bus_service_increase/scripts/warehouse_queries.py#L16-L81) -- no major change here except removing `collect()` so that the joins happen without looping through individual operators, but can do it for weekday/sat/sun (3x).
* [analysis dataset, part 1](https://github.com/cal-itp/data-analyses/blob/bus-service-workflow/bus_service_increase/scripts/create_analysis_data.py#L11-L186)

tiffanychu90:
@edasmalchi: Been going through some sanity checks. I want to rerun and generate some more of the later parquets (`shape_frequency`. `shape_frequency_funding`, `shapes_initial`, etc), but leaving some thoughts here before I go an change those.

1. `sanity-check-A1.ipynb` traces the differences all the way back to `shape_frequency` file. 
1. `sanity-check-shape-frequency.ipynb` walks through some of it.

* Going back to your runtime calculation, because it handles the trips that start in/around midnight correctly.
* But, will simplify the logic of keeping middle stop / finding the mean within the departure hour. Essentially, rather than going back to the full dataset (`st_trips_joined`), stay with just keeping the middle stop (`middle_stop` df) for each `trip_key`. This allows each `trip_key` to be represented once in any average taken. With the full dataset, the same `trip_key` can appear multiple times, and would throw off the average.  (https://github.com/cal-itp/data-analyses/blob/c8aa5693fc230552d9e9296273c95f1dd59f0407/bus_service_increase/scripts/create_analysis_data.py#L82-L109)
* One kind of edge case:  `trip_keys` whose 1st stop occurs in a different departure hour than the middle stop, but that's fine, eventually, that trip is binned to the departure hour of its middle stop, by virtue of only keeping the row for trip's middle stop.

edasmalchi:
@tiffanychu90

- Great, just keeping the middle stop sounds good to me
- I looked at it for a while and I'm also unsure why `shapes_processed` is so different in row count
- ... but, as you might have noticed, `shape_frequency` is quite different too (~95,000 rows in new vs ~423,000 in old). My original code created rows with 0 trips for hour for all departure hours x shape_ids where that shape_id wasn't running at all. It's important to account for this, since we want to be able to calculate increases for a particular shape at a particular hour even if there is _currently_ no service. (`A3_service_increase_estimator` relies on this). I've copied some code over from A1 that accomplishes this by first generating a multiindex of all possible shape_id/day_name/departure_hour, then setting that as the index for the single-operator shape_frequency df. That results in NaNs for departure_hours where no service currently exists, then just an easy .fillna(0) turns those rows into useful data. (though there's also a need to refill id columns and it could be refactored somewhat)
- Let me know if you'd like me to work that logic into the scripts, or if you think it's better handled another way

```
    ## insert nulls or 0 for missing values as appropriate
    shapes_routes = shape_frequency[['shape_id', 'route_id']].set_index('shape_id').to_dict()['route_id']
    iterables = [shape_frequency.shape_id.unique(), ['Thursday', 'Saturday', 'Sunday'], range(0, 24)]
    multi_ix = pd.MultiIndex.from_product(iterables, names=['shape_id', 'day_name', 'departure_hour'])
    shape_frequency = shape_frequency.set_index(['shape_id', 'day_name', 'departure_hour'])
    try:
        shape_frequency = shape_frequency.reindex(multi_ix).reset_index()
    except ValueError: ## aggregate rare shape_ids with multiple routes (these may be errors)
        duplicate_indicies = shape_frequency.index[shape_frequency.index.duplicated()]
        print(f'''
        caution, operator has {len(duplicate_indicies)} duplicate shape/day/departure_hour \
out of an ix of {len(shape_frequency.index)}
        ''')
        assert len(duplicate_indicies) < len(shape_frequency.index) / 100, 'too many duplicate shape/day/departure_hour'
        shape_frequency = shape_frequency.groupby(['shape_id', 'day_name', 'departure_hour']).agg(
            {'calitp_itp_id':max, 'route_id':max, 'trips_per_hour':sum, 'mean_runtime_min':np.mean})
        shape_frequency = shape_frequency.reindex(multi_ix).reset_index()

    shape_frequency['calitp_itp_id'] = (shape_frequency['calitp_itp_id']
                                        .fillna(method='bfill')
                                        .fillna(method='ffill')
                                       )
    shape_frequency['trips_per_hour'] = shape_frequency['trips_per_hour'].fillna(0)
```

tiffanychu90:
@edasmalchi Ahaha, ok, that makes sense. I had done a check where if I dropped your NaNs, our observations differ by much less, and they're in the same ballpark. So I wasn't too, too worried, and with the runtime calculation restored the way it was, I think I'm happy with this portion. 

Can you work that logic back into the script, but separate it out as a function (just because this is an explicit step to build out the df to be what it needs for the analysis), with a little description like you just explained above?

The tract categorization stuff was good to go, just slow (because of clipping), and I tried it with `gpd.overlay` + `pd.pivot` to go reshape it from long to wide, and it only shaved off a little in the time, so, I'd rather just keep the code as is. I don't know if there's a way to reduce how computationally expensive that is.

edasmalchi:
Just committed zero generation, should slot into the create_analysis_data.py flow OK

tiffanychu90:
Status:

* for tract-level, data creation through viz is done. Clean up of old util files done.
* for service estimator, data creation through `A1` is done, `A2` is off in a separate `create_tract_categorization.py` script, needs to be worked into the `create_analysis_data.py`

PR #69, opened 2022-02-08
Round2 bus service

tiffanychu90:
Continue migrating `bus_service_increase` stuff from notebooks A2, A3 into scripts.

Previous PRs: #63 
GH issue: https://github.com/cal-itp/data-infra/issues/1003


tiffanychu90:
@edasmalchi: This [is where A3 stuff moved to](https://github.com/cal-itp/data-analyses/blob/round2-bus-service/bus_service_increase/setup_service_increase.py). Checks are good, NTD datasets get pulled the same way. I stashed one more parquet before the `increase_by_operator` one to get the full one.

This [sanity check notebook](https://github.com/cal-itp/data-analyses/blob/round2-bus-service/bus_service_increase/sanity-check.ipynb) goes through and compares the `shapes_initial` dataset differences. I'm puzzled about why we each miss some operators, although now the observations are much closer, but any aggregation differences can be traced back to here, although I don't have any more ideas for how to debug (beyond having a way to get `shapes.txt` for the actual date of service, and not the most current...but I'm not sure if we would ever have a `dim_shapes` version)

PR #76, opened 2022-02-15
Consolidate MSD

tiffanychu90:
Instead of displaying static altair images, display and allow for limited interactivity, hover and tooltip, line selection.  

See how that displays when using `papermill` and `nbconvert` to get HTML.

Previous PRs: #31, #40, #41, #44, #51, #57, #59, #62


tiffanychu90:
@edasmalchi: I can give your notebooks this treatment, but will wait until the branch `more-msd-tweaks` is merged in. Some of the stops/trips queries we both use are fairly similar, save for a couple extra columns, so I consolidated those so the entire MSD dashboard is thought of more holistically.

PR #83, opened 2022-02-23
analyses(pre-commit): Added pre commit hooks, black formatting, gh workflow to shared utils

charlie-costanzo:
This PR puts in place black pre-commit hooks to check and fix formatting in the _shared_utils directory

Files added:
.pre-commit-hooks.yml


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,14 @@
+name: pre-commit
+
+on:
+  pull_request:
+  push:
```
atvaccaro, line 5: this might not be a bad idea to run on every push, i.e. per commit on a PR

does this run just once on a PR right now?

atvaccaro (approved):

.github/workflows/pre-commit.yml
charlie-costanzo: yep it appears to be running on every commit on a PR @atvaccaro

PR #84, opened 2022-02-23
MSD workflow

tiffanychu90:
Migrate more of 01 into `create_coverage_data` into scripts, do sanity checks.

Previous PRs: #31, #40, #41, #44, #51, #57, #59, #62, #76, #81


tiffanychu90:
@edasmalchi: Can you take a look at [this notebook](https://github.com/cal-itp/data-analyses/blob/21eabc411e5a54e9d0a86ae3d578ea436e5ac1d3/msd_dashboard_metric/01-new.ipynb) and the functions here to create the datasets [in this script](https://github.com/cal-itp/data-analyses/blob/21eabc411e5a54e9d0a86ae3d578ea436e5ac1d3/msd_dashboard_metric/create_coverage_data.py)?

The block accessibility metrics are the same. But, there are slight differences in the RT tract stats, see print statements before the **Summarizing Coverage** section. Very slight...has to do with I think how blocks/tracts ended up getting joined, duplicates dropped. The polygon-on-polygon (block to tract) join kept making the kernel restart, so I stuck a crosswalk / `pd.merge` to be a quicker operation.

If you're happy and ok with these, I'll move it into your existing 01 notebook, and also stash the intermediate parquets in GCS.

edasmalchi:
@tiffanychu90 looks good to me, thanks!

PR #129, opened 2022-04-15
DLA Updates

natam1:
- updates to dla_utils.py and gen_report.py adding HTML for chart titles and removing charts from tab widget
- incorporating changes into charting function work notebook
- add a test notebook for reports calling function in gen_report.py


natam1:
now on PR#185

PR #134, opened 2022-04-18
Deploy analyses to multiple individual sites

atvaccaro:
Adds:
1. Top-level index site with links to individual analyses
1a. This provides a single URL and the other sites are redirects underneath
2. Splits up deploys separately, so sites can be worked on more independently
3. Templates the table-of-contents and config for each analysis site, reducing boilerplate

Still TODO, likely in a follow-up:
1. Support more than 1 notebook per site (technically this should already work with part/chapter-level notebooks, but should test with an actual analysis)
4. Support unparameterized notebooks (this works but still has section dropdowns under a chapter, vs 1 file per chapter)


edasmalchi (approved):

atvaccaro:
re-opening momentarily

PR #186, opened 2022-04-19
Revert "Rt selfserve 2"

edasmalchi:
Reverts cal-itp/data-analyses#122


edasmalchi:
thought I somehow got rid of Andrew's script but I think I've found it in the portfolio folder, no need to revert after all

PR #189, opened 2022-04-20
make portfolio index deployable in prod

atvaccaro:
Currently, the portfolio index is only built on PRs to a preview URL, for example https://data-analyses-189--cal-itp-data-analyses.netlify.app. This PR adds a workflow case to run on `main` and deploy to the actual production index site, cal-itp-data-analyses.netlify.app.

### DIFF
--- a/.github/workflows/update-portfolio-index.yml
+++ b/.github/workflows/update-portfolio-index.yml
@@ -2,7 +2,7 @@ name: Update and deploy portfolio index site
 
 on:
   push:
-    branches: [master]
+    branches: [main]
     paths:
       - '.github/workflows/update-portfolio-index.yml'
       - 'portfolio/**'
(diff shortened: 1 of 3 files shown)


.github/workflows/update-portfolio-index.yml
```diff
@@ -1,6 +1,11 @@
 name: Update and deploy portfolio index site
 
 on:
+  push:
+    branches: [master]
```
lauriemerrell, line 5: Should this be main?

lauriemerrell:
And can you add a PR description?

.github/workflows/update-portfolio-index.yml
atvaccaro: yes ty!

lauriemerrell (approved):

PR #246, opened 2022-05-18
Add CPI

natam1:
- adding CPI package to _shared_utils


tiffanychu90 (commented):
Can you cherry pick the `precommit.txt` commit? I think this will allow the check to pass. If you run those in the terminal, then do `git add`, `git commit`, it will run those precommit checks, and then you can look with `git status` if those hooks modified anything. If they did, you need to do `git add` and `git commit` again 

https://github.com/cal-itp/data-analyses/tree/add-cpi (OOPS, used a branch name too similar to yours...you can delete the branch once you add that file)

PR #258, opened 2022-05-27
Format DLA Reports

natam1:
- move report utils to dla_utils 
- format outputs with new functions in dla_utils
- rerun reports


natam1:
note: getting an error when rerunning parameterized reports

Issue #226, opened 2022-05-09
Epic - Parallel Corridors to State Highway Network


tiffanychu90:
After receiving a research request, use this template to plan and track your work. Be sure to also add the appropriate project-level label to this issue (eg gtfs-rt, DLA).

### Summary
* Which transit routes are considered parallel to the State Highway Network (SHN)?
* Which transit routes should be targeted for further service improvements?
* Which highway corridors should be targeted for new service?

### Research required:
* Define "parallel to the SHN" - what criteria, thresholds of % overlap of transit route and/or % overlap highway counts as parallel?
* From all the parallel corridors and routes, narrow it down to the top ___ corridors for potential investment
* Of these transit routes, which ones have GTFS RT? For operators, identify ____ routes. Pair it with RT analysis, show where the bottlenecks are occurring on those routes.

### Notes, misc:
* Follow-up to bus-service-increase work [here](https://github.com/cal-itp/data-analyses/issues/163) and [here](https://github.com/cal-itp/data-analyses/issues/164)
* Depends on [assembled routes / stops](https://github.com/cal-itp/data-analyses/issues/160)
* Aligns with GTFS RT work

### Reviewers [Stakeholders]

1. Chad Baker
2. Planning & Modal Advisory Committee

## Research Issues

- [x] 1. #155  

### Past PRs
#56, #58, #64, #96, #104, #123, #131, #193, #199, #207, 

#### Issues
* #217 
* #224 
* #225 

#### PRs
* #203 
* #214 
* #270
* #272 

### Deliverables
* [PMAC presentation 3/9/22](https://docs.google.com/presentation/d/1ohRf3X6Lms52uTqYZ_rDt25Cl1mMPjRkzj2fFo92Vyk/edit)
* [Portfolio](https://analysis.calitp.org/parallel_corridors/README.html) in favor of presentation below
* [Parallel Corridors presentation](https://docs.google.com/presentation/d/1ZSxAFwz0NtYJpcDzjx4rlZ4Rdyv2baB_DJPo50pDFjM/edit?usp=sharing)

PR #270, opened 2022-06-03
Tweak parallel corridors

tiffanychu90:
Add some tweaks based on feedback:
* Larger dots on the stripplot for readability
* Use `route_short_name` instead of `route_id`
* Exclude operators who are only showing 1 route in each route_group from having a page. Also exclude chart if an operator only has 1 viable competitive route within the route_group. More about UI/UX...confusing to the reader because they think the results are cut-off. Showing at least 2 viable routes makes it less confusing.

[Epic GH issue](https://github.com/cal-itp/data-analyses/issues/226)


tiffanychu90:
Close because too many merge conflicts when building the jupyterbook tests

PR #276, opened 2022-06-07
Build with 2 notebooks as chapters

tiffanychu90:
Test Jupyterbook build with 2 different notebooks as chapters. This case might be useful for similar topics (HQTA), where we've been adding unparameterized, single notebooks for stakeholders. Stakeholders might have interest in other lines of work related to that topic?

Incorporate @edasmalchi's SACOG Green Means Go HQTA map into this build. Sits alongside Tier 1 facilities in HQTAs.

```
# Test ways to add static images:
# Method 1: 
![image](https://raw.githubusercontent.com/path-to-image)

# Method 2: 
<img src="./some-image.png", alt="image", width="200px">

Does not work: ![image](./some-image.png)
```


tiffanychu90:
To do: @edasmalchi: I reran and saved the `hqta_green.html` and set `legend_dict=None` (which was something new I added). That link you gave Abby Jackson won't break, but the new link created in the portfolio is: https://analysis.calitp.org/hqta/0__hqta_green_SACOG__.html. You can unstage the `hqta_green.html` if you want later.

PR #280, opened 2022-06-08
Parse XML for ESRI metadata

tiffanychu90:
* Import XML metadata that was manually filled out
* Convert to json to use a dictionary to replace certain values
* Cobble together automation steps of updating metadata, part within ArcGIS using `arcpy`, but some in Hub, because it's way more usable


tiffanychu90:
@j-meelah: Almost there with this PR. [Script](https://github.com/cal-itp/data-analyses/blob/parse-xml/open_data/metadata_update.py).

PR #290, opened 2022-06-14
Where is Metrolink (323)?

tiffanychu90:
Debug why Metrolink (ITP ID: 323) is missing from `ca_transit_routes`. Merges are working as expected, but maybe this is not what we intend?

Dig into [warehouse queries](https://github.com/cal-itp/data-analyses/blob/main/traffic_ops/prep_data.py#L55-L69) for trips to see what's happening with `shape_id`

### DIFF
--- a/traffic_ops/missing-metrolink.ipynb
+++ b/traffic_ops/missing-metrolink.ipynb
@@ -1,0 +1,877 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "fdcdf3f1-2c34-4749-9496-5e1a1f6e0692",
+   "metadata": {},
+   "source": [
+    "# Check why Metrolink (ITP ID: 323) is missing from routes shapefile"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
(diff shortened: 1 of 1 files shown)


tiffanychu90:
@lauriemerrell: Can you check what's feeding into the `views.gtfs_schedule_dim_trips` table for 6/21/22? I do some digging into my queries [in this notebook](https://nbviewer.org/github/cal-itp/data-analyses/blob/d041045a59fa459a12522937386d6bcb0ebf1bde/traffic_ops/missing-metrolink.ipynb) and seems like all the `shape_id` values are `None` in that view.
```
dim_trips = (tbl.views.gtfs_schedule_dim_trips()
             >> filter(_.calitp_itp_id==323)
             >> distinct()
            )
```

lauriemerrell:
Ok this is a raw data problem -- they do *provide* `shapes.txt` but they don't actually provide `shape_id` in their `trips.txt` 😅 All their shapes are flagged as unused in their validation output

image

image

I'm not totally sure where that leaves you analysis-wise but I believe that is what is going on

tiffanychu90:
Closing for now - no error in code assembling the `routes` file for open data. Raw data issue.

PR #311, opened 2022-07-13
Cleaning up 5311 work.

amandaha8:
Cleaned, organized, and improved the code for 5311 written back in February. The new script to be referenced is "[data_prep_refactored.py](https://github.com/cal-itp/data-analyses/blob/cleanup/5311/data_prep_refactored.py)"


tiffanychu90:
@amandaha8:

* Update the link above to read from the `main` branch, since once it's merged in, that link (off the working branch) is broken.
* Switch out hard-coding values for denominator and use something like `df[df.year==base_year].column_name.iloc[0]`: https://github.com/cal-itp/data-analyses/blob/main/5311/data_prep_refactored.py#L112
* Fix this, consistent naming of columns, now that you're going to re-run script (fix notebooks so charts using `automobile_doors` column also works): https://github.com/cal-itp/data-analyses/blob/main/5311/data_prep_refactored.py#L287
* Is this correct? `df1` is overwritten twice, and `df2` means different things here: https://github.com/cal-itp/data-analyses/blob/main/5311/data_prep_refactored.py#L349-L351
* Use `on='name'` instead of `left_on`, `right_on`, since column name is the same: https://github.com/cal-itp/data-analyses/blob/main/5311/data_prep_refactored.py#L349-L351
* Method chain as much of this as you can: https://github.com/cal-itp/data-analyses/blob/main/5311/data_prep_refactored.py#L357-L366
* Indent properly for readability: https://github.com/cal-itp/data-analyses/blob/main/5311/data_prep_refactored.py#L383-L395
* Great!!! https://github.com/cal-itp/data-analyses/blob/main/5311/data_prep_refactored.py#L414-L419

amandaha8:
@tiffanychu90 I have a question about the bullet point: indent for proper readability. Can you clarify what I should do? Thanks! https://github.com/cal-itp/data-analyses/blob/grant_5311/5311/data_prep.py#L391

tiffanychu90:
@amandaha8: Indenting for readability example: https://github.com/cal-itp/data-analyses/blob/main/_shared_utils/shared_utils/calitp_color_palette.py#L6-L13=. The `shared_utils` were passed through a formatter, and the 2nd and subsequent lines are all indented. Instead of having them all left aligned to the edge, they should have some indents, even if there are multiple items on the 2nd, 3rd, 4th line.

Issue #252, opened 2022-05-23
Feature Request: CSS templates 


tiffanychu90:
**Where does your feature apply?**
Select from the below, and be sure to affix the appropriate label to this issue (e.g. `dataset`, `jupyterhub`, `metabase`, `analysis.calitp.org`)
- [ ] Data (the warehouse)
- [ ] JupyterHub
- [ ] Metabase
- [ ] analysis.calitp.org  
- [x] Other (add detail)

**Is your feature request related to a problem? Please describe.**
Currently, rendered GH pages uses `jekyll-minimal`, which looks pretty bare compared to `analysis.calitp.org`.  

**Describe the solution you'd like**
Add a CSS template for GH pages look more like analytics.calitp.org or reports.caliltp.org.

**Describe alternatives you've considered**
analysis.calitp.org uses a `sphinx-rtd` theme. I'm not tied to that theme in particular. It seems like CSS styling would be the easiest to sit on top of existing GH pages, such as [this](https://docs.calitp.org/data-analyses/bus_service_increase/img/highways-no-parallel-routes.html) and [this](https://docs.calitp.org/data-analyses/consolidated_app/lessons_learned.html). 

Right now, styling is done within a notebook. Beyond using the Cal-ITP color palette for charts, there's no much standardization elsewhere.

Any alternative that applies a template over HTML and Markdown pages within the `data-analyses` repo will work!

**Additional context**
Epic #250. This is a feature request / ticket for Ryon's work.

amandaha8:
When printing out a data frame, it would be great to have style it with a scrollbar, remove the gray-white banding (at least IMO) and center  the text.

tiffanychu90:
Not sure if we want some similarities to how our analysis.calitp.org looks, especially for styling of tables / fonts. Here's an [example](https://analysis.calitp.org/parallel_corridors/district_07-los-angeles/23__competitive-parallel-routes__district_07-los-angeles__itp_id_182.html) of the basic `sphinx-rtd` theme Jupyterbook uses.

ryon:
Hi @tiffanychu90, all of the links to GH pages in this issue seem to be 404ing for me. Do I need access, or are there any new links to look at? I could use some good targets for styling. Thanks!

tiffanychu90:
@ryon: Thanks for catching that...I didn't realize someone turned off GH pages. All the links should be working now! Some are very plain, some have attempted at adding very simple stuff like colored line breaks.

[...]

PR #312, opened 2022-07-14
Set up basic workflow and theme for GitHub Pages

ryon:
Per https://github.com/cal-itp/data-analyses/issues/252, this sets up a workflow for local testing of GitHub Pages rendering, and adds some bare-bones layouts and CSS to make rendered pages look a bit more like `reports.calitp.org`.

Current limitations / room for improvement:

- Adding the Jekyll/GH Pages infrastructure to the root directory in `main` may not be ideal
- Jekyll is grabbing every random .md file it can find and turning it into a "page" (maybe this is desired?)
- Layout is more-or-less mobile responsive but minimal-effort
- Large tables are prone to overflowing in a not-so-nice way, might require wrappers or other trickery to get them under control
- There is no code highlighting or other enhanced features


ryon:
Sample screenshots of rendered pages:

Screen Shot 2022-07-14 at 4 07 17 PM

Screen Shot 2022-07-14 at 4 07 10 PM

ryon:
Update: I've excluded `portfolio` in the `_config.yml`, but I'm not sure there's a way to include full HTML files in the render process. What would you want to happen with them? If they got shoved into the Jekyll layouts, they'd end up with duplicate `<html>` tags and whatnot.

The index is fully customizable by editing either `home.html` (the layout) or `index.markdown` (which gets rendered and inserted into the `{{ content }}` block of the layout). I modified the layout to collapse the full list of pages, and the markdown file to demonstrate how we could hand-curate some links.

tiffanychu90:
@ryon: I would want this [notebooks](https://docs.calitp.org/data-analyses/bus_service_increase/img/highways-no-parallel-routes.html) or [this notebook](https://docs.calitp.org/data-analyses/consolidated_app/lessons_learned.html) to have better styling than what was there with the `jekyll-minimal` theme. These html pages are the primary use case for the styling...rather than the Markdown READMEs. The Markdown files are a nice bonus, but even if they had no additional styling, they already serve their purpose as is, whereas the analyses rendered as html are shown to stakeholders.

PR #377, opened 2022-08-18
PMP Script 

amandaha8:
- Improved efficiency. Instead of stacking dataframes from the previous accounting periods, read them directly from an Excel workbook that will hold all previous months. 
- Changed monetary columns from int to float for accuracy.
- Script now contains the cleaning function, so run only one function to create the entire Excel workbook.


tiffanychu90:
@amandaha8: the [function](https://github.com/cal-itp/data-analyses/blob/ac8c55dded7df48534ac5e206f8399c1f2001ccd/pmp_dashboard/A1_utilities.py#L439-L445) you want to call when you run it, you would set the variables that you're using as arguments. Basically, `file_name`, `name_of_sheet`, ...etc all need to have values assigned to them, or else the function won't run. 

Something like this (using caps to distinguish between what's a variable with a value assigned and what's the argument you set in the function,/but hasn't taken on a value yet): 
```
if __name__=="__main__":
    SPREADSHEET_FILE_PATH = "my_excel.xlsx"
    my_dict = {
       "sheet1": sheet_1_list_to_filter,
       "sheet2": sheet_2_list_to_filter,
    }
    ACCOUNTING_PERIOD = 2
    YEAR = "2022"
    for sheet_name, lists_to_filter in my_dict.items():
        pmp_dashboard_sheets(SPREADSHEET_FILE_PATH, sheet_name, lists_to_filter, ACCOUNTING_PERIOD, YEAR)
          
```

PR #378, opened 2022-08-23
Update Netlify deploy key

thekaveman:
We're migrating to a single org-wide Netlify deploy key. 

This PR just updates the name of the secret used for the Netlify deploy.


thekaveman:
Seems like the new key works!

atvaccaro (approved):

PR #395, opened 2022-08-29
Kmk ahsc tblbuild

KatrinaMKaiser:
AHSC analytical table:
- Added geometry
- Time buckets - Weekday am peak, middday, pm peak; Sat 6a-7p, Sun 6a-7p
- Operators: LA Metro and SB MTD
- Next steps: identify (filter?) rail stops and drop


hunterowens:
FYI I haven't looked at the LA Metro data in years but it was bus only AFAIK

KatrinaMKaiser:
@hunterowens this PR deals with stops, not ridership,  #393 looks at the ridership :)

Issue #398, opened 2022-08-30
Research Task - Make sure HQTA points/polygons data is correct


tiffanychu90:
## Research Task

Use this template to create issues that represent small chunks of work (.5 - 2 days) that stand alone, or represent incremental work as part of an [Epic Issue Template](https://github.com/cal-itp/data-analyses/issues/new?assignees=&labels=epic&template=epic.md&title=Epic+-+%5BResearch+Request+Project+Title%5D).

* **Relevant PR** (if applicable):  

* **Question or Goal**: **BUG**: @edasmalchi noticed in the published open data portal, the CA HQ points datasets has incorrect points. [Slack thread](https://cal-itp.slack.com/archives/C03JL54P8QY/p1661822418098349) 

* Double check July data to make sure it's ok, and push that through metadata rewrite and overwrite what's published.

* **Data Required**:
  * Currently have:
  * Need:  

* **Research Required**:  

* **Metrics**: what specific calculations are you looking to find?  

* **Expected Outputs / Findings**: what do you hope to have created when this is complete?

edasmalchi:
@tiffanychu90 Actually the points dataset [CA HQ Transit Stops](https://data.ca.gov/dataset/ca-hq-transit-stops1) seems ok, rather it's the polygon dataset [CA HQ Transit Areas](https://data.ca.gov/dataset/ca-hq-transit-areas1) that has all the major stops on what looks like a single point/small buffer. Apologies for being unclear on Slack.

PR #402, opened 2022-09-01
July open data

tiffanychu90:
Build on PR (#400) and step through rest of workflow to prepare for open data portal.

## Traffic Ops datasets
* Run July traffic ops (`ca_transit_routes`, `ca_transit_stops`) 
* **Fix** `portfolio_utils.latest_itp_id()` to correctly filter in `siuba` with `_.calitp_itp_in_latest == True`, and not `_.calitp_itp_in_latest is True`. Otherwise, it's returning an empty df, and paring down `ca_transit_stops` to zero rows.
* End up caching `stop_times` to compare which `stops_ids` are present on a day, instead of using `views.gtfs_schedule_index_feed_trip_stops`
* TODO: figure out where to store cached files, it's used in `rt_delay` (indiv operators), `traffic_ops` (compiles indiv operators + add's Amtrak (ITP ID 13) back in, and `bus_service_increase`. Maybe the compiled datasets should live in the same folder somewhere so we can better track when it's used across projects

### HQTA datasets
* **Fix**`hqta_details` is missing in `ca_hq_transit_areas` after the dask rewrite. Buses are split into ones `hq_corridor_bus` and others, drawn with separate buffers, and concatenated. Add `hqta_details` in for `hqta_type==hq_corridor_bus` so it's present after the concatenation, instead of holding NaNs
* TODO: add notebook to do map checks in the workflow

## Open Data Workflow
* Take July HQTA and traffic ops and run through metadata steps for ESRI -- add `open_data/Makefile`
* Check HQTA points and polygons don't show up in Stanislaus National Forest (they're good!)
* Prepare file gdb for emailing to Caltrans GIS 
* Update `intake_justification.md` to run through steps for opening a ticket / filling out form / attaching email with supervisor approval / description of justification for why this is public data / attaching sharepoint URL

Previous PR: #400
GH issue: #398


tiffanychu90:
@hunterowens @edasmalchi Opened an open data portal ticket to push an update to existing layers (all 4 datasets) this morning. I don't believe there's an email sent to me if it's updated, so we'll have to be on the lookout.

Issue #302, opened 2022-06-30
Epic - Estimating Ridership Increases from Transit Investments (AHSC)


edasmalchi:
### Summary
* The [Affordable Housing and Sustainable Communities](https://sgc.ca.gov/programs/ahsc/) (AHSC) program is a unique grant  funding housing and transit projects using funds from California's Cap-and-Trade program.
* Estimating actual greenhouse gas reductions from a proposed transit investment is important to operating the program successfully, but there is no universally applicable methodology or tool at present.
* Our partners at the [California Strategic Growth Council(https://sgc.ca.gov/) and [California Air Resources Board](https://ww2.arb.ca.gov/) have existing tools to estimate GHG reductions from _ridership_ increases, so if we're able to develop something workable for estimating ridership increases from various kinds of transit investments that would be a great help.
* We should be able to use GTFS Schedule data, plus demographic and land use data, to inform what sort of ridership gains we might expect if a specific operator makes investments to increase frequency, speed, span of service, etc...

### Research required:
* Which evidence-based methodologies can we find in practice or literature to support developing this work?
* How can we operationalize these?
* How can examples of application information provided serve as meaningful inputs to whatever models we can develop?
* More TBD

### Notes, misc:

> Thanks again Hunter, Tiffany and Eric for your help with this. To recap, you will help us develop a way to estimate ridership increase based on a funded transit service increase. We would ideally release some type of public online tool in October/November along with other application materials.

> I’m attaching back up documentation for the six service type improvements we most commonly fund. These letters are also what we use to verify inputs from transit agencies, so any recommendations for improving this would be helpful

[47610 - GHG Transit Component--MTS Capital Improvement.pdf](https://github.com/cal-itp/data-analyses/files/9023882/47610.-.GHG.Transit.Component--MTS.Capital.Improvement.pdf)
[47693 - GHG Transit Component--MTS CNG Bus Purchase.pdf](https://github.com/cal-itp/data-analyses/files/9023883/47693.-.GHG.Transit.Component--MTS.CNG.Bus.Purchase.pdf)
[47700 - GHG Transit Component--Capital Corridor Long distance bus.pdf](https://github.com/cal-itp/data-analyses/files/9023884/47700.-.GHG.Transit.Component--Capital.Corridor.Long.distance.bus.pdf)
[...]

PR #405, opened 2022-09-07
AHSC ACS EDA

KatrinaMKaiser:
Load, filter, join ACS variables to census tract boundary shapefile
Part of #302


KatrinaMKaiser:
Todo: only keep variables of interest, create derived variables

Issue #399, opened 2022-08-30
Research Task - Put August HQTA dataset into ESRI enterprise geodatabase


tiffanychu90:
## Research Task

Use this template to create issues that represent small chunks of work (.5 - 2 days) that stand alone, or represent incremental work as part of an [Epic Issue Template](https://github.com/cal-itp/data-analyses/issues/new?assignees=&labels=epic&template=epic.md&title=Epic+-+%5BResearch+Request+Project+Title%5D).

* **Relevant PR** (if applicable):  

* **Question or Goal**: For Aug update, let's do it through the enterprise geodatabase. With the Microsoft ODBC driver installed and the connection to the enterprise geodatabase established, figure out rest of workflow for pushing a new table. What's metadata update process look like?

* **Data Required**:
  * Currently have:
  * Need:  

* **Research Required**:  

* **Metrics**: what specific calculations are you looking to find?  

* **Expected Outputs / Findings**: what do you hope to have created when this is complete?

Issue #321, opened 2022-07-22
Epic - Daskify HQTA


tiffanychu90:
After receiving a research request, use this template to plan and track your work. Be sure to also add the appropriate project-level label to this issue (eg gtfs-rt, DLA).

### Summary
* HQTA workflow currently relies on running several notebooks each month.
* Move existing work from notebook into scripts, and also incorporate `dask` and `dask_geopandas` to shorten the run time from several hours to...something way less.
* Refactor the code to make it more modular within each script. Also break apart steps into distinct stages so it shows how we translate an ambiguous and complicated policy definition into data processing stages.
* Retain some notebooks to double check visualizations at various stages 
* There's some overlaps with `rt_delay` work, so move over the utilities functions related there into `shared_utils` to be `rt_utils`, etc. 
* Research how to move `dask` computations to the cloud, once it runs locally successfully, to make full use of the partitioning. 
* **Update: 8/23/22 discussion with Jameelah**: we're not ready to spin up a dask cluster yet, since HQTA itself is not a compelling enough use case to justify Jarvus resource cost to spin it up. If the RT work, more of the spatial processing needed, can use `dask`, that would make the cost justifiable. We need to move existing Python code into using `dask` first and will revisit this discussion when we've done so.

### Research required:
* How to use `dask`, `dask_geopandas` 
* Reconstruct the logic to use the more limited operations in `dask_geopandas`, and use `geopandas` if we really can't 

### Notes, misc:
* 

### Reviewers [Stakeholders]

1. @edasmalchi 

## Issues

- [x] 1. #313 - initial rewrite / grouping into scripts of work present in notebooks. mostly data download, data cleaning, processing.
- [x] 2. #320 - reconceptualize using `dask_geopandas.sjoin` first to narrow down rows to do `dask_geopandas.clip`. new step to cut down computing time.
- [x] 3. #332 - step through rest of workflow to create final `hqta_points`, `hqta_areas` geoparquets
- [x] 4. #353 - refine download data step to use `gtfs_utils`. accommodate the fact that different operators might be "valid" each month, and enforce the conditions in which all the merges should be successful (so later steps don't break down)
- [x] 5. #358 - refine, instead of picking longest `shape_id` by route_length, figure out the symmetric_difference and pick longest `shape_id` in each direction, combine into 1 row, and then cut that one line into segments.
- [x] 6. #400 - debug error (#398) of points in a forest. run through a full hqta workflow for July and Aug, add logs to better debug
[...]

PR #411, opened 2022-09-08
Aug hqta

tiffanychu90:
### HQTA

* **Fix**: `hqta_areas` takes in dissolved `hq_corridor_bus` before drawing new buffer around it. Dissolve to `calitp_itp_id-route_id` level.
* Save extra intermediate outputs in GCS from `C4_create_bus_hqta_types`, so building final outputs is faster 
* Add in `compile_operator_data.py` to take in compiling / caching functions that was in `traffic_ops/prep_data`
* Add in `route_id` to both `hqta_points` and `hqta_areas`
* `shared_utils.utils`: add a `geojson_gcs_export` function and take `geojson` or `geojsonl`
* Add exports to `geojson` and `geojsonl` too
* Add notebook checks to do some map sanity checks: `check1_downloads`, `check2_hq_corridors`, `check3_hqta_points`

### Traffic Ops
* Move workflow of compiling all the operator files to HQTA, since HQTA needs it for adding `route_id` info to `stop_id` (requires `trips` and `stop_times` as intermediary tables)
* Run Aug for `ca_transit_stops` and `ca_transit_routes` -- put this into enterprise geodatabase to see how metadata update goes 

Epic: #321 
GH issue: #398 , make sure this update makes it for #399


tiffanychu90:
@edasmalchi: For HQTA areas, do you want the `hqta_type == hq_corridor_bus` to be dissolved by `calitp_itp_id-route_id`? `route_id` isn't kept in the final export, multiple rows would be present for the same attributes (`itp_id`, `hqta_type`, `hqta_details`), but different geometries. Would this be confusing? 

If it's too unwieldy to have only 1 dissolved area for each operator, then maybe adding back `route_id` for all the other `hqta_types` would be useful too?

This is a non-issue for HQTA points, since the `stop_id` is present to distinguish between multiple rows.

edasmalchi:
I think it would be ideal to have a bit more granularity than the operator level. Agree that it could get confusing to have identical-seeming rows with different geometries.

So maybe adding back `route_id` for all the other `hqta_types` is the way to go?

tiffanychu90:
@edasmalchi: Can you check the `hqta_points` and `hqta_areas` files? Added `route_id` for both files, since there's `stop_id`. Would rather not have any nulls or else it's extra confusing.

I opted not to add the `route_name` since that should be available in the `ca_transit_routes` also in the open data portal, and that's the one that's pulling a standardized name `route_name` (same as what's published in the portfolio)

```
import geopandas as gpd
from utilities import GCS_FILE_PATH

points = gpd.read_parquet(f"{GCS_FILE_PATH}hqta_points.parquet")
areas = gpd.read_parquet(f"{GCS_FILE_PATH}hqta_areas.parquet")
```

edasmalchi:
@tiffanychu90 These look good to me!

One small thing, we seem to have calculated hqtas for a small part of Yuma, AZ. Should we go ahead and clip to CA at some point in the process?

yuma

tiffanychu90:
For Yuma (ITP ID 386):
* HQTA points: `gpd.sjoin` with `predicate="within"` vs `predicate="intersects"` doesn't get rid of it for points 
* HQTA polygons. `predicate="within"` would filter out Yuma

After discussing, would rather have consistency between the HQTA points and polygons datasets. All points should get buffered and show up in the polygons dataset.

PR #412, opened 2022-09-09
AHSC Stop Table Modification

KatrinaMKaiser:
Issue: #302 
Today (9/8):
- joined route_type, saved out intermediate parquet. 
- Filtered to LA Metro only (still waiting on SB MTD data)


KatrinaMKaiser:
Todo: make additional derived acs vars, put back stop geometry, acs eda

Issue #370, opened 2022-08-16
Epic - Recommendations for Better Buses


tiffanychu90:
After receiving a research request, use this template to plan and track your work. Be sure to also add the appropriate project-level label to this issue (eg gtfs-rt, DLA).

### Summary
Come up with a list of priority 100 worst buses / segments. Provide a suite of highest priority / worst offenders, covering intersections, routes, highways with no parallel or competitive transit service. Give each district a prioritized to-do list.

[Slack thread](https://cal-itp.slack.com/archives/C02H6JUSS9L/p1660253072696009)

#### Existing Routes
* Combine work related to competitive parallel routes + speedmaps -> back out route-level df showing % trips competitive (or lack thereof), average speed (mph), delay at endpoint, whether route is parallel or not
* Score across % trips competitive, average speed - equally weighed, let's surface worst offenders (relative to all routes), see what comes up
* Pick the 50-75% of recs from this. Can find where the bottlenecks are happening, depending on the route.

#### Highways - Dearth of Service (Zero Parallel Transit or Not Competitive)
* These will likely target more rural areas where there's long stretches of highway with no service
* There won't be too many from this? Maybe 20ish?
 
#### Highway Corridors with Transit...can be improved
* Find transit routes that run mostly on the SHN, deeper dive into whether these are competitive or not?
* May include new analysis? Let's see what surfaces

### Notes, misc:
* 

### Reviewers [Stakeholders]

1. Hunter Owenes
2. Jeannie Ward Walker - exec board presentation in Sept?

## Issues

- [ ] 1. #
- [ ] 2. #

### Deliverables
* Fill out exec board template by Fri 8/19/22
* Some notebooks with deeper dive analysis
* Maybe new portfolio site of worst offenders? Portfolio site for each district, and show for that district, all their existing operator routes that could be improved, highway segments that have nothing, and corridors that benefit from new stuff that affect multiple operators

tiffanychu90:
cc #494 tracking additional development

PR #418, opened 2022-09-13
Highway segments with speeds

tiffanychu90:
* Calculate trip weighted mean speed mph to plot for highway segments in `G5_plot_segments`
* Re-order logic of scripts: assemble route-level stats in `G2`, assemble highway segment level stats in `G4`
* `altair_utils`: point at `styleguide` and use Cal ITP color palette and chart styling function, 1 more step toward deprecation
* `portfolio_utils`: missing ITP IDs in `tbl.airtable.california_transit_organizations`, add what district they're in, and include in the `add_caltrans_district` function
* PMAC follow-up: add a map to plot routes by `parallel` / `on_shn` / `other` in `D5_plot_routes_shn`

Epic: #370


tiffanychu90:
Note:  `G3_calculate_speeds_all_operators.py` relies on cached `rt_delay/rt_trips` operator files for 5/4/21, compiled into 1 parquet.

PR #435, opened 2022-09-20
Update readme

natam1:
- add readme for 5310/5311 analyses


natam1:
@amandaha8 feel free to change/add if needed!

PR #437, opened 2022-09-22
More AHSC updates

KatrinaMKaiser:
Issue: #302 
This branch is primarily for adding SBMTD data received earlier this week, but also some smaller intermediate changes
- Saved out an LA Metro only version of analytical file for regression exploration
- added sbmtd to trips table so far
- in process - ingesting ridership - need to figure out how to join on non-standard stop IDs/stop codes/stop names? 
- todo: add SB county ACS stuff


KatrinaMKaiser:
Issue: negative stop ids are not matching stop codes in warehouse data

TODO Approach:
- inner join on stop code
- anti-join both datasets to get remainders
- join remainders on name (levenshtein edit distance or just change "/" to " & ")
- concatenate

KatrinaMKaiser:
Finished processing ridership. Threw out 6 stops where ridership data couldn't be matched.
Next branch: load SB area ACS data

PR #438, opened 2022-09-22
Project Priority

amandaha8:
- First look through the data to see how many null values, any duplicated project names, etc.
- Created fake data for a mock up of a parameterized notebook.
- First stab at creating a parameterized notebook  for every project, worked on graphs, charts, and markdown cells.


tiffanychu90:
@amandaha8:

* nice job on the dual bar chart and pie chart functions!
* is an extra space needed after [return](https://github.com/cal-itp/data-analyses/blob/b8cbc9aec793467c4f9b1ea39b2ae6290295db62/project_prioritization/_utils.py#L93)?
* know the difference between [this](https://github.com/cal-itp/data-analyses/blob/b8cbc9aec793467c4f9b1ea39b2ae6290295db62/project_prioritization/_utils.py#L14-L20) and ```

PR #450, opened 2022-09-28
Add2report

natam1:
- add districts to analysis report 
- add and explore May 4 table (with all operators)
- filter down calitp_url_numbers for agencies with multiple


natam1:
Closing without merging because I cherry-picked commits to new branch

PR #460, opened 2022-09-30
Sep hqta

tiffanychu90:
* Run 9/14/22 for open data
* Run through HQTA and Traffic Ops 
* Move ArcGIS Desktop script to ArcGIS Pro script (mostly ability to use Python functions, f-strings, and some of the `arcpy` syntax has changed
* metadata update process still occurs within JupyterHub....and there are now some parts commented out, such as bounding box info, etc that used to be in XML export with ArcGIS Desktop, but isn't present in XML export with ArcGIS Pro
* can write to enterprise gdb directly...each layer takes around 6-7 min

TODO: explore whether the metadata class in `arcpy` can actually add all the FGDC kind of metadata....seems like from [these arcpy docs](https://pro.arcgis.com/en/pro-app/latest/arcpy/metadata/metadata-class.htm) that you can only update title, description, but not any of the other required info

GH issue: #449


tiffanychu90:
Submitted open data request form

Issue #472, opened 2022-10-07
Research Task - function to easily view/generate rt_delay data


edasmalchi:
## Research Task

* **Question or Goal**: Analysts should be able to easily find out if rt_delay intermediate data is already present on a given date, and if not have the option to generate it using rt_analysis.OperatorDayAnalysis and rt_analysis.OperatorDayAnalysis.export_views_gcs() without using or modifying rt_delay/04_generate_all.ipynb

* **Expected Outputs / Findings**: A function (in shared_utils.rt_utils?) that enables this sort of process

* **Notes**: See initial work at end of https://github.com/cal-itp/data-analyses/blob/main/rt_delay/23_sccp.ipynb

KatrinaMKaiser:
new util `get_operators` in `shared_utils.rt_utils`

PR #478, opened 2022-10-10
Util for checking/generating OperatorDayAnalysis intermediate data products

KatrinaMKaiser:
Issue #472 

- opened nb to test function
- if it works for other stakeholders, will put function in rt_utils

### DIFF
--- a/_shared_utils/shared_utils/rt_utils.py
+++ b/_shared_utils/shared_utils/rt_utils.py
@@ -727,41 +727,46 @@ route_type_names = {
     "12": "Monorail",
 }
 
+
 def get_operators(analysis_date, operator_list, generate_new=False):
     """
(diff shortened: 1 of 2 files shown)


edasmalchi:
Looks good! 

One suggestion: perhaps instead of/in addition to printing the function could return results in a dictionary, which could make it a bit easier to integrate

Maybe something like `{'already_ran': list_already_ran, 'not_yet_run': list_not_yet, 'newly_run': list_newly_run}`

edasmalchi:
> The small change I would make: using `generate_new is True` as opposed to `generate_new == True`. I think for boolean `is` and `is not` is more pythonic than `==`?

Actually the most pythonic might be simply `if not generate_new:` and `if generate_new:`

Here's what [pep8](https://peps.python.org/pep-0008/) has to say about it:

Screenshot 2022-10-11 091004

tiffanychu90 (approved):

Issue #491, opened 2022-10-14
Bug: fix `rt_utils` import of `rt_analysis`


tiffanychu90:
**Where did the bug occur?**
Select from the below, and be sure to affix the appropriate label to this issue (e.g. `dataset`, `jupyterhub`, `metabase`, `analytics.calitp.org`)
- [ ] Data (the warehouse)
- [ ] JupyterHub
- [ ] Metabase
- [ ] analytics.calitp.org  
- [x] Other (add detail)

**Describe the bug**
Figuring out the right path for importing `rt_analysis`? @edasmalchi, do you also get this when you import the packages from within `rt_delay`?

**To Reproduce**
1. In notebook: `from shared_utils import rt_utils`
Error: 
```
ModuleNotFoundError                       Traceback (most recent call last)
Cell In [2], line 1
----> 1 from shared_utils import rt_utils

File ~/data-analyses/_shared_utils/shared_utils/__init__.py:1
----> 1 from . import (
      2     altair_utils,
      3     calitp_color_palette,
      4     geography_utils,
      5     gtfs_utils,
      6     map_utils,
      7     portfolio_utils,
      8     rt_dates,
      9     rt_utils,
     10     styleguide,
     11     utils,
     12 )
     14 __version__ = "0.1.2"
     16 __all__ = [
     17     "altair_utils",
     18     "calitp_color_palette",
   (...)
     26     "utils",
     27 ]

File ~/data-analyses/_shared_utils/shared_utils/gtfs_utils.py:15
     13 import siuba  # need this to do type hint in functions
     14 from calitp.tables import tbl
---> 15 from shared_utils import geography_utils, rt_utils, utils
     16 from siuba import *
     18 GCS_PROJECT = "cal-itp-data-infra"

File ~/data-analyses/_shared_utils/shared_utils/rt_utils.py:23
     20 from siuba import *
     22 sys.path.append("../../rt_delay")  # noqa
---> 23 import rt_analysis as rt  # noqa
     25 # from zoneinfo import ZoneInfo
     26 # import warnings
     28 fs = gcsfs.GCSFileSystem()

ModuleNotFoundError: No module named 'rt_analysis'
```
Works in a notebook within `shared_utils`, but not in a notebook in another directory like `high_quality_transit_areas`
**Expected behavior**

**Additional context**

edasmalchi:
ah, sorry about this! I tested it from `rt_delay`, where it does work.

It can be "fixed" in other project folders with
```
import sys
[...]

PR #492, opened 2022-10-17
Temp disable rt util

edasmalchi:
- Actually moving to rt_analysis.py as a short-term fix


edasmalchi:
fixes shared_utils import from #491

PR #532, opened 2022-11-04
AHSC Spatial Model Finalize

KatrinaMKaiser:
Related issue: #302 AHSC Epic

Added n_routes to the stop-level analytical file, reflecting that transfer stations may have more ridreship. Need to explore strength of this relationship and add to regression.

Also planning to remove spatial lag of ridership, add spatial lags of independent variables maybe? Need ways to describe spatial relationships between stops that don't depend on ridership (since we won't have it from applicants)


KatrinaMKaiser:
- Added n routes, spatial lag of pop density, non-citizen, and poverty.
- Changed spatial relationship between nearby stops to square root distance decay within 400m (about a quarter mile)
- Bottom line: about 2% increase in stop-level ridership for every additional bus trip serving a stop during a weekday

PR #546, opened 2022-11-11
AHSC Big Table(s?)

KatrinaMKaiser:
Issue #302 (AHSC grant ridership model)

Goal: Create giant table for all bus operators and all bus stops, with trips per weekday/saturday/sunday, routes per weekday/saturday/sunday, and route short names stored somewhere.

Keep running out of memory. Have tried:

- previous workflow, using `gtfs_utils `for trips/routes/stops/stoptimes
- using dask df options for `get_stop_times` (could research this more)
- pulling from `gtfs_schedule_data_feed_trip_stops_latest`  

Next step: may iterate over operators, save multiple parquets (since expected applicant is only 1 operator at a time)


KatrinaMKaiser:
Successfully queried all operators, intermediate data in GCS bucket. Need to concatenate, add spatially-lagged regressors. Also still need to generate saturday/sunday model coefficients.

Issue #564, opened 2022-11-29
Research Request - migrate `gtfs_utils` to v2


tiffanychu90:
Complete the below when receiving a research request, and continue to add to this issue as you receive additional details and produce deliverables. Be sure to also add the appropriate project-level label to this issue (eg gtfs-rt, DLA).

## Research Question
**Single sentence** description:  Move `gtfs_utils` queries that pulls from v1 in warehouse to v2
  
**Detailed** description: Test this on Sept and Oct dates we already have downloaded for HQTA to make sure feed -> operators is done correctly for summing up service hours statewide

## How will this research be used?
What will be achieved or informed? What is the intended impact?

## Stakeholders & End-Users
Who is asking the question, and who is the audience?

## Metrics
* any specific calculations
* you think will be needed

## Data sources
* **Cal-ITP data sources**:

* **External data sources**:

* **Remaining data source questions**:

## Deliverables:
Notebook, dashboard, narrative, presentation details.

## Timeline of deliverables:
Estimated completion date

tiffanychu90:
* Patch needed here for v1 to fully work with existing analyses #628

tiffanychu90:
* Adding more to `portfolio_utils` for v2 in #626

PR #570, opened 2022-12-05
Migrate gtfs utils

tiffanychu90:
### `gtfs_utils`
* Set up initial `gtfs_utils_v2.py` to mimic what was done in `gtfs_utils`, but leave out the subsetting columns and subsetting by other columns
* Rework some of the logic, no need for subsetting, but need some new tables. We want to move a lot more quickly as dbt models:
   a. Expand `get_trips` in v2 to encompass `get_route_info`, since these are typically used together
   b. Deprecate `get_route_info` then? Leave in script for now, but don't move into dbt
   c. `get_route_shapes` is straightforward move, simply update the `make_linestring` function to use dask to apply it. Add a count of how many trips occurred for this `shape_id` that might be useful.
   d. `get_stops` is straightforward move, also add in `route_type` 
   e. new table: `feed_key` to `organization name` table created so analysts have option to subset by `organization name` or `feed_key`
* TODO: think more about how `get_stop_times` would be joined / subsetted / aggregated. Need to make the Metrolink fix that was in `gtfs_utils` for `get_trips`.
* #564

### DIFF
--- a/_shared_utils/shared_utils/dbt_for_gtfs_utils_v2.sql
+++ b/_shared_utils/shared_utils/dbt_for_gtfs_utils_v2.sql
@@ -1,0 +1,230 @@
+/*
+Move gtfs_utils_v2 into SQl to sit as dbt model to create another
+mart_* table that analysts start with.
+
+-- NEW TABLES TO ADD (rename everything if necessary) --
+1. fact_feeds (rename to something else?)
+    Purpose: see what feed_keys - organization names are present on selected day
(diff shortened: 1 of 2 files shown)


tiffanychu90:
New `mart_gtfs` tables created in and query-able with `tbls.mart_gtfs.fct_*` in:
* https://github.com/cal-itp/data-infra/pull/2123
* https://github.com/cal-itp/calitp-py/pull/95

Issue #494, opened 2022-10-17
Epic - 100 Routes to Better Buses


KatrinaMKaiser:
### Summary: This epic is for tracking PRs/smaller issues on this deliverable
* Leverage existing GTFS-RT work (speedmaps, comparisons with SHN) to deliver ~10 recommendations to each district
    * Transit Corridors
    * Intersection/Spot treatments
    * SHN segments with transit service opportunities

### Notes, misc:
*  [Exec Board Presentation](https://docs.google.com/presentation/d/1EnmYkq7ebESxLqctWEdc0nt81pheH660zIktB0SRF4c/edit#slide=id.)

### Reviewers [Stakeholders]

1. @tiffanychu90 
2. @edasmalchi 
3. @KatrinaMKaiser 

## Issues

- [ ] 1. #441 (portfolio deployment)
- [ ] 2. #

### Deliverables
* Parameterized notebook [site](https://analysis.calitp.org/one-hundred-recs/README.html)
* Spreadsheet of recommendations, for tracking action
* shapefiles/geoJSON for districts to use in planning

edasmalchi:
may reboot as part of the SB125 work!

edasmalchi:
on-system we've learned a lot through the transit priority work
off-system, cover via other SB125 issues

edasmalchi:
giving ourselves more credit

PR #601, opened 2023-01-11
Better Buses: District 2

KatrinaMKaiser:
Related issue: #494 (Better Buses notebook site corridor+hot spot analyses)

Where we are leaving off:
- Started on D2, `better_bus_utils.get_sorted_transit_routes` returned 2 operators with no GTFS-RT
- Need to bring on districts for operators missing them

Next week:

- @edasmalchi will resolve this using either `shared_utils.portfolio_utils.add_caltrans_district` or the district from Airtable as part of a flexible utility function taking district and corridor/hot spot as parameters
- Will need to rebase when I come back


KatrinaMKaiser:
left off: 

- D2 is tabled - will handle manually since only operator does not have district appropriately tagged
- Started D6 with helper function; only 1 operator (GET Bus) will have relevant corridors

Issue #441, opened 2022-09-23
Portfolio Request - 100 Recs


tiffanychu90:
Complete this issue template when you are ready to deploy parameterized or standard notebooks to the analytics portfolio website.

### Notebook(s) to be parameterized and/or deployed
* TBD

Epic: #370 
Quick version with unparameterized notebooks: #442 

### Select as many as needed from the below
- [ ] Unparameterized notebook
- [ ] One notebook deployed to own site
- [x] Multiple notebooks deployed into own site 

### Identify parts / chapters / sections
For more information on how to structure your notebook, visit the Jupyter Book documentation for [Structuring the Table of Contents](https://jupyterbook.org/en/stable/structure/toc.html).  

Part = 100 Recs
Chapter = District
Section:  4 notebooks
(1) speedmaps for transit routes major improvements
(2) speedmaps for marginal improvements
(3) highway corridors with transit on SHN
(3) highway corridors with no transit

**Ex 1**: RT analysis / district / ITP ID  
**Ex 2**:  district / ITP ID / notebook A, notebook B

* Part
  * Chapter
    * Section

tiffanychu90:
* this PR only launches test portfolio related to #494

PR #658, opened 2023-02-22
More bb param

edasmalchi:
- parameterize remaining manually identified corridors/hotspots for odd-numbered districts


edasmalchi:
#441

Issue #689, opened 2023-03-16
Research Request - Pedestrian Crash Fatalities among People Experiencing Homelessness


KatrinaMKaiser:
## Research Question
To what extent are people experiencing homelessness (PEH) represented among KSI crashes on the SHN? 
  
Caltrans PMP is interested in improving pedestrian safety on its system. People experiencing homelessness are a particularly vulnerable population. Would like to statistically test the correlation between KSI crashes on the SHN and the presence of unhoused people. 

## How will this research be used?
Will use to identify priority locations for pedestrian safety upgrades, homelessness outreach?

## Stakeholders & End-Users
Working with Abby Jackson (PMP) and the Division of Safety

## Metrics
EDA metrics:
* K/SI per overall pop (density) per tract touching SHN 
* K/SI per unhoused pop (density) per tract
* correlation coefficients/significance between K/SI and unhoused pop density

Longer term metrics (need more data):
* K/SI per job density per tract (lower priority) 
* Percent of deaths among PEH due to motor vehicle crashes, by tract or other aggregation geography - use to validate
* Percent of overall K/SI on SHN that are PEH
* Utilizing mode share and VMT to generate crash rates as opposed to counts

## Data sources
* **Cal-ITP data sources**:

* **External data sources**:
    * SWITRS data
    * CT Encampments data

* **Remaining data source questions**:
    * We'd need coroner data focused on homelessness specifically?
    * Can we leverage count data aggregated up by census tract put out by counties/CoCs? 

## Deliverables:
Easy deliverable: ArcGIS webmap or 1-page notebook site
Longer term deliverables: story map or parameterized notebook site with maps (a page for each county/Continuum of Care for which we have data)

## Timeline of deliverables:

- [x] Interim check-in
- [x] Planning meeting
- [x] Presentation: April 19

KatrinaMKaiser:
Did EDA in ArcGIS Pro today. Steps:

1. Loaded relevant data: Encampments, SWITRS (filtered to ped crashes), SHN, Bridges
2. Projected all layers
3. Buffered bridges 200m (short walking distance)
4. Summarized crashes and encampments within buffers
5. Used "Local Bivariate Relationship" tool to identify bridges (buffered) where there's a significant relationship between encampments and ped crashes (based on 30 nearest neighbors)

![image](https://user-images.githubusercontent.com/47205245/227064372-1ecc5ec2-39de-48f9-b646-eb1c0e31ac86.png)

Lessons learned: 

[...]

PR #697, opened 2023-03-24
Homelessness Data Processing

KatrinaMKaiser:
Relevant issue: #689 

Cleaning and aggregating relevant data for analysis:

1. Create study areas from buffered bridge data ✅ 
2. Filter crashes to pedestrian crashes, aggregate K/SI to study areas
3. Aggregate encampment cleanup work orders to study areas, including counts, normalized values (percentiles?), and a dummy variable approach


KatrinaMKaiser:
TODO after Thursday meeting: Update scripts to use pyesridump, test script run

KatrinaMKaiser:
pyesridump will work for bridges (on open data portal) but may not be reliable for internally hosted data (encampments).

Met w/ Abby today and discussed large number of D11 records. Some of these may just be inspection or trash pickup work orders miscoded as encampment cleanup. Need to add a step filtering out non-homelessness related rows. Also want to do EDA of which districts tend to use which words.

KatrinaMKaiser:
All data has been processed, next step is to join the aggregated encampments with the bridge data and write a makefile ordering the scripts. Next week: go back to do that after some EDA on statewide joined data

PR #701, opened 2023-03-29
Put shared_utils requirements back

tiffanychu90:
* undo removing `shared_utils/requirements.txt` from #696 
* this https://github.com/cal-itp/data-infra/pull/2418 did not actually install those packages, so add it back in


atvaccaro:
ah sorry @tiffanychu90 this is my bad, I failed to update the image version in the Jupyterhub deploy to point at the new image; will fix as part of my dbt dev environment work https://github.com/cal-itp/data-infra/issues/2438

Issue #605, opened 2023-01-11
Research Request - Initial road segments (Census/Caltrans LRS)


tiffanychu90:
Complete the below when receiving a research request, and continue to add to this issue as you receive additional details and produce deliverables. Be sure to also add the appropriate project-level label to this issue (eg gtfs-rt, DLA).

## Research Question
**Single sentence** description:  Cut Census roads into segments
* #592   

**Detailed** description: Explore how road segments will tie into routes, trips, shape_ids. Do we need an order of segments the way we have for route segments?

## Metrics
* Include all primary and secondary roads, whether or not they intersect with GTFS shapes.
* Only include local roads that intersect with GTFS shapes. 
* Buffer the original Census Bureau file first to find the intersection, then bring the original geometry back in. 
* For GTFS shapes, start with stops with intersecting to local roads -> buffer around 50 feet. Filter out for linearids that are already found. 
* Then move onto roads -> buffer. 
* Cut all the secondary, primary, and intersecting local roads into segments. 

## Data sources
* **Cal-ITP data sources**: 
* GTFS shapes. 
* GTFS trips.

* **External data sources**: 
* Census Tiger- all California roads are located in GCS bucket in the shared_data folder as "all_roads_2020_state06.parquet".

## Deliverables:
* Script and parquet file. 

## Timeline of deliverables:
* ASAP.

PR #739, opened 2023-04-28
Tiger

amandaha8:
- First draft function that cuts all roads with stops and routes. 
- Dissolved tiger roads on linearid, so there is only one row for one road.


tiffanychu90:
* #605

PR #777, opened 2023-06-06
Refresh portfolio

tiffanychu90:
Fix dead links...why does netlify expire sites hosted? 

Refresh:
* `atp`: also update `calitp` to `calitp_data_analysis` and remove `altair_utils` references
* `ahsc`
* `competitive_corridors`:  remove link to RT because can't determine order in URL if order is not same order in `rt.yml`
* `dla`: also update `calitp` to `calitp_data_analysis`

Remove:
* `segment_speeds`: test segment website
* `one_hundred_recs`: outdated


tiffanychu90:
* cc #622 last time big refresh needed for no reason

PR #799, opened 2023-07-03
CSIS Examples for Deck

KatrinaMKaiser:
Preparation for July workshops: Maps and example scores for:

- [x] Safety metric
- [x] Land Use/ Natural Resources metric
- [ ] Freight??
- [ ] VMT??


KatrinaMKaiser:
Will come back to Land Use with:
- version that gets stops 
- Rural example

VMT will have typologies, not specific projects for now
Freight still TBD

PR #812, opened 2023-07-20
add new gtag to analytics site

atvaccaro:
I dropped the ball on this; we've had an outdated google analytics tag configured for the analysis site.

Note: sites will need to be regenerated for this to take effect.

### DIFF
--- a/portfolio/templates/index.html
+++ b/portfolio/templates/index.html
@@ -18,6 +18,13 @@
             line-height: 1.2
         }
     </style>
+    <script async src="https://www.googletagmanager.com/gtag/js?id=TAG_ID"></script>
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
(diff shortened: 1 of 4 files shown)


portfolio/templates/index.html
```diff
@@ -18,6 +18,13 @@
             line-height: 1.2
         }
     </style>
+    <script async src="https://www.googletagmanager.com/gtag/js?id=TAG_ID"></script>
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date());
+      gtag('config', 'G-JCX3Z8JZJC');
```
thekaveman, line 26: Is it possible to reference the ID from the yaml file?

thekaveman (approved):

portfolio/templates/index.html
atvaccaro: good call, I templated it in everywhere via Jinja

Issue #708, opened 2023-04-07
Docs: Analyst Starter Kit


tiffanychu90:
*Reference*: [docs.calitp.org/data-infra/](https://docs.calitp.org/data-infra/)

Most of the docs should make it to our current docs page using the `data-infra` repo. 

Use this GH issue to track PRs related to **adding new notebooks / tutorials in `data-analyses`** that are referenced in `docs.calitp.org`. 

Brainstorming [docs](https://docs.google.com/document/d/1NIQq841WSl3l-vu6yHmn-zE6cN5KIoi5x3hSGvtTv_I/edit#heading=h.1ce8fwl3f4gd) here.

**What type of changes need to be made?**
Select all that apply
- [x] Edits
- [x] Additions

**Describe the changes**
### `data-infra` docs
* [x] @edasmalchi: intro to GTFS, either editing / expanding existing docs, add supporting notebook(s), and/or create slide deck with assigned reading
* [x] @natam1: intro to grants, include flow chart. [Metabase tutorial module](https://docs.google.com/document/d/12B71T1rppLBv8_7-QoKdKd9Bn8NdzEGxPvu-pyfnmqc/edit)
* [x] @amandaha8: intro to warehouse, include most frequently used tables for `trips`, `shapes`, `day` summaries. Include at least 1 table each major `mart` category (`mart_gtfs`, `mart_gtfs_quality`, `mart_transit_database`). Update docs with new links to practice exercises / update old links (folder path moved), add more resource links (pull from practice exercise notebooks). 
* [x] @KatrinaMKaiser: Acronym list, updating docs overall with onboarding schedule
* [x] @tiffanychu90: scripts for future-proofing our work
### `data-analyses`
* [x] Instead of `data-analyses/example_report/`, move relevant practice exercises and new exercises into `starter_kit/`
* [x] @tiffanychu90: add more practice exercises and cover `geopandas` and `dask`.
* [ ] @edasmalchi: include GTFS tutorial, show examples of what's in `routes`, `shapes`, `trips`, `stop_times`, etc

PR #848, opened 2023-08-23
add gtfs schedule table demo

tiffanychu90:
* drop demo of `trips`, `shapes`, `stops`, `stop_times` into `starter_kit`
* #708


tiffanychu90:
@amandaha8: use this to teach pull from `origin/main` and getting "new files" you need

PR #854, opened 2023-08-28
Low speeds

### LINKED ISSUE #804: Research Request - Fix Certain Speeds

amandaha8:
* Fixes #804 
- The speed maps looked slow (in particular, the Big Blue Bus) and this PR addresses the speeds through:
- Calculating the length of the segments we have on file.
- Removing segments that don't have RT data. 
- Dividing meters lapsed by said segment length.
- Filtering out rows in which the meters lapsed is less than 55% of the recorded segment (for now, this can be adjusted) when calculating the peak and all-day averages. 
- Add the cut rows back in after the averages are calculated.
- Includes a graph that visualizes the percentage of rows cut by percentile after filtering by segment length.


rt_segment_speeds/scripts/B2_avg_speeds_by_segment.py
```diff
@@ -108,23 +137,29 @@ def speeds_with_segment_geom(
...
-            "district", "district_name"
-        ]
-    ).to_crs(geography_utils.WGS84)
+    # Grab left only results, which did not show up in the stats df
+    missing = missing.loc[missing._merge == "left_only"].reset_index(drop = True)
+    missing = missing.drop(columns = ['_merge'])
+    
+    # Concat & clean.
+    stats2 = pd.concat([missing, stats])
+    stats2 = stats2.fillna(stats2.dtypes.replace({'float64': 0.0, 'object': 'None'}))
```
tiffanychu90, line 154: What are the `stats2 = pd.concat / fillna` steps doing here? Instead, if in line 142, you do 2 left merges of segment geom with `all_day` and then `peak`, that would get you the same result

rt_segment_speeds/scripts/B3_export.py
```diff
@@ -99,29 +99,6 @@ def grab_extreme_tails(
...
 
-def suppress_bad_data(
-    gdf: gpd.GeoDataFrame, 
-    quantile_cutoff: float = 0.05
-) -> gpd.GeoDataFrame:
-    """
-    Suppress the tails of bad data that we need to investigate.
-    Suppress the bottom 5% of extreme calculations.
-    """
-    cutoffs = grab_extreme_tails(
```
tiffanychu90, line 110: If the `grab_extreme_tails` is not used, can you also remove that function too?

tiffanychu90 (changes requested):

tiffanychu90:
Two more general comments: 
1. Can you rebase off of `main` and force push to remove Christian's commit? 
2. Can you expand the PR description and link relevant GH issues?

rt_segment_speeds/scripts/B2_avg_speeds_by_segment.py
amandaha8: That was a way to bring back the unique segments that were dropped and concat them with the stats dataframe. The fillna is to replace any of the segments that were dropped with 0's instead of nulls and none instead of peak/all day.

So I should do something like this? 
`
stats = pd.concat([
        all_day.assign(time_of_day = "all_day"),
        peak.assign(time_of_day = "peak")
    ], axis=0)

segments = segments.to_crs(geography_utils.WGS84)

gdf = pd.merge(
        segments,
        stats,
        on = SEGMENT_IDENTIFIER_COLS,
        how = "left"
    )
`

tiffanychu90: Yes, I prefer left joins. Filling with zeros might mean that `df[df.speed_mph==0]` grabs a completely different set of rows from `speeds_{date}` and `avg_speeds_{date}`. If there are zeroes in speeds, then somehow removing them and averaging them still results in a row with 0's...that could be confusing.

tiffanychu90 (approved):

Issue #852, opened 2023-08-24
Research Request - SHS mainline delay to match signal metrics


edasmalchi:
## Research Question
  
> Can we get the delay at the mainline i.e. what segments of the SHS have the worst reliability when it comes to transit? 

**Detailed** description:
#783 follow-up from Traffic Ops

## Metrics
* SHS filtered and concatenated segment speeds (from speedmap pipeline)

## Timeline of deliverables:
Estimate 8/25 - 8/28

edasmalchi:
reopening to track work on a potential non-spatial version per Traffic Ops ask

PR #858, opened 2023-08-30
Patch muni speedmap

edasmalchi:
- download a new vehicle positions date for August since the old one was missing Muni
- rebuild speedmap site
- revisit signal analysis to export matching full State Highway System speeds data per #852


edasmalchi:
kept both sets of dates and logs (@tiffanychu90 and I inadvertently picked separate rerun dates for August)

Issue #785, opened 2023-06-15
Research Request - use vehicle positions near stop as actual_stop_arrival


tiffanychu90:
Complete the below when receiving a research request, and continue to add to this issue as you receive additional details and produce deliverables. Be sure to also add the appropriate project-level label to this issue (eg gtfs-rt, DLA).

## Research Question
**Single sentence** description:  Find GTFS RT vehicle positions nearest stop to get actual stop arrival.
cc #719 
  
**Detailed** description: 
As the next phase of schedule vs RT work: Implement interpolation of vehicle positions nearest stops. Pick one of the two approaches:
* last position prior to a given stop
* last position prior to a given stop and first position after the stop, average the timestamps (done currently in speedmaps).

## How will this research be used?
Use RT to back out a modified `stop_times` table with actual stop arrivals (vs schedule stop arrivals in GTFS schedule). Can be used in Conveyal.

## Metrics
* actual stop arrival for each trip-stop

## Notes
* include a stage to pare down vp
* functions used to take at least 2 trips across 2 operators

## Data sources
* **Cal-ITP data sources**:
* vehicle positions on a single day
* schedule `stop_times`...use this as template for operator-trip-stop columns. generate the RT corresponding `arrival_time` column as `actual_arrival_time` 
* GitHub folder: `rt_schedule_v_ran` and GCS folder: `rt_vs_schedule`

edasmalchi:
goal: interpolate using before/after points as in `rt_analysis`, but first filter the universe of vehicle positions to a more manageable level

PR #933, opened 2023-10-24
Find vp nearest stop

tiffanychu90:
## segment speeds
* expand `map_partitions` usage more...create a `shapely_project_all_vp.py` script. We can project the vp position along shape_geometry fairly quickly...and use this to visualize loop_or_inlining shapes all at once
* once we have each vp's `shape_meters`, combine that with the prepped `stops_projected` to get stop positions. 
* for each trip in the vp dataset, to finding the nearest vp to each stop position for each trip (`loop_or_inlining==0` only)...not sure how to work on more complex shapes yet
* attach the nearest vp and subseq vp and get the interpolated arrival time (set up the same way as in `rt_utils`)
* set up `24_plot_interpolation_results.ipynb` to see what the results look like
* actually, this basically provides the RT derived `stop_times` table needed for #785
   * except, all the `loop_or_inlining==1` trips are missing from this right now


tiffanychu90:
@edasmalchi: some early work to see stop arrivals using interpolation: `f{SEGMENT_GCS}stop_arrivals_{analysis_date}.parquet")`, only for the sep2023 date.

Issue #885, opened 2023-09-19
Research Request - Bus Service Increase 2023-24


edasmalchi:
## Research Question
**Single sentence** description:  
Rebuild #164 for v2 warehouse. Also generate and catalog new insights.

**Detailed** description:

- [ ] How many routes meet service target by operator?
- [ ] Where are there missing routes?
- [ ] How can GTFS data inform service coordination and/or consolidation, especially in urban areas?

## Deliverables:
Notebook, dashboard, narrative, presentation details.

## Timeline of deliverables:
Estimated completion date

tiffanychu90:
* Include costs and estimates of increase in ridership (tweak model from AHSC, Katrina / Shweta's work)

tiffanychu90:
1/30/24: SB 125 mtg update. We can continue to use NTD 2019 values for `cost_of_bus`. Reproduce the service density vs CalEnviroScreen tracts (but swap CalEnviroScreen -> RHNA by tracts) to identify where there's enough density to support more service but doesn't have enough service

tiffanychu90:
Closing because dataset was given to McK related to this

PR #937, opened 2023-10-27
V2 bus service increase

tiffanychu90:
* Move `bus_service_increase` analysis we did 2 years ago from v1 warehouse into v2 warehouse
* Refactor much of the setup and make use of our cached parquets and aggregate those / merge those together
* Remove most the CalEnviroScreen by tract charts/maps and leave 1 chart, 1 map in the notebook
   * Rather than update some of the scripts and functions, we can remove everything that is unused. Too many `altair` saved pngs, `folium` htmls, we can use `chart.interactive()` now and `gdf.explore()` 
* Archive 100 recommendations and competitive-routes scripts into `v1_scripts/` folder
* TODO: much of the NTD metrics that are used live in saved parquets or `views.transitstacks`....didn't found corresponding tables in current warehouse...a lot of those tables are deprecated currently. Figure out where / if these exist in `mart.transit_database`. This [bridge table](https://dbt-docs.calitp.org/#!/model/model.calitp_warehouse.bridge_organizations_x_funding_programs) has funding program name (5307/5311) but not dollar amts. 
   * Need to create 1 singular script to clean these....consolidate the existing work out there that exists in several folders and set that to be most upstream 
* #885


tiffanychu90:
For pre-commit hooks, suppress 2 rules that we can't meet (space after semicolon...`gs://` and multiple statements need semicolon).

Tried to set the github action setup and checkout from v3 to v2, but that didn't help with passing CI. Revert that and just deal with some of the flake8 comments where we can.

PR #938, opened 2023-10-27
precommit linter issue, set python version

tiffanychu90:
Do the same in `data-analyses` as in `data-infra` 
* https://github.com/cal-itp/data-infra/pull/3054


tiffanychu90:
For pre-commit hooks, suppress 2 rules that we can't meet (space after semicolon...gs:// and multiple statements need semicolon).

Tried to set the github action setup and checkout from v3 to v2, but that didn't help with passing CI. Revert that and just deal with some of the flake8 comments where we can.

PR #974, opened 2023-12-19
initial route variance identification work

csuyat-dot:
initial work related to 924


tiffanychu90:
Notes here for continuing work on this issue:
* don't use `fct_monthly_routes` for this, compile across multiple dates using `helpers.import_scheduled_trips`.
* differences we're interested in (if we have 4 dates):
   * `route_id`: check `nunique(gtfs_dataset_key-route_id combination) > 1`. 
   * if `nunique == 4`, then route_id is changing every month, if `nunique == 2`, then route_id changes once in this period
   * check `nunique(gtfs_dataset_key-route_long_name-route_short_name) > 1`
* pick a couple examples to show where this is true for all operators

Issue #966, opened 2023-12-14
Research Request - Trip Level to Route Level Metrics 


amandaha8:
## Research Question
**Single sentence** description:  
* After completing #936 comparing rt versus scheduled data on a trip level, aggregate each trip up to the route level to get an overview.  
* Relates to  #824
  
**Detailed** description: TBD. 

## How will this research be used?
Understand how each route performs in terms of gtfs data density, timeliness, etc. 

## Stakeholders & End-Users
Caltrans 

## Metrics
* TBD
## Data sources
* Metrics housed here: calitp-analytics-data/data-analyses/rt_vs_schedule/trip_level_metrics

## Deliverables
* Script
* Notebook

## Timeline of deliverables
TBD

PR #987, opened 2024-01-05
Route Aggregation

amandaha8:
- This is my first attempt at aggregating up the trip level statistics to route level based on Tiffany's work.
- I experimented with `groupby` to see if there was any difference with observed=True, observed=False, etc and try to make things faster.
- I wrote a function that checks the original trip level rows with the aggregated one row=one route results because some of the groupby combos yielded mistakes.
- I edited Tiffany's original code to exclude columns that are no longer needed/relevant.
- I added in new columns for the metrics derived from the trip level statistics such as how many minutes of the trip recorded 2+ GPS pings. 
- I reran all the trip level statistics for the months because some columns are added later in the route aggregation step.
- This PR does not address trips that are outliers. For example, there are some trips in which the scheduled trip time is only 12 minutes but the difference between the first and last time stamp is 1000 minutes.


tiffanychu90:
* #966

PR #1004, opened 2024-01-22
added .env.sample, changed mtc pull

vevetron:
Some minor changes to the checking script so that it pulls and parses the xml directly, instead of the xlsx method.


amandaha8:
Wow this is great, thanks for updating it! You can merge it in! @vevetron

PR #1022, opened 2024-02-13
Dgs bus cost

csuyat-dot:
Continuing work related to issue #897

Created new `dgs_usage_report_bus_analysis` notebook to read in and clean DGS usage report data from Rebel. Then moved code to `dgs_data_cleaner` python script. Ran script and ensured data was exported to GCS as parquet file.

Afterwards, refactored both the `FTA_bus_grant_analysis` and `tircp_bus_analysis` notebooks and moved code to separate python scripts. (`fta_data_cleaner` and `tircp_data_cleaner` respectively). Both scripts ran and exported parquet files to GCS.

Updated `cost_per_bus_analysis` notebook to read in new parquet files. Then refactored/moved cleaning steps to `cost_per_bus_cleaner` script, leaving the exploratory code in the notebook.

Next, refactor and rerun all the exploratory code with the new dataset. Tighten up scripts as needed.


tiffanychu90:
* The skeleton for how functions should be constructed is there. Now wrap up the steps into functions. Functions start with a `def some_name():` and return something at the end. Don't leave the steps in-line in the long form.
* There are some variables that are defined at the top. If you see them used later, call the variable and use it! -- This is true for all the scripts you set up. If you define your path `GCS_FILE_PATH` once, you can call it in other scripts with `from dgs_data_cleaner import GCS_FILE_PATH` and use the same variable elsewhere.
   * Ex: `GCS_FILE_PATH` or `GCS_PATH` is defined once at the top. Later, when you export, don't write out the full path again, use f-strings to construct your export file path (`f"{GCS_FILE_PATH}some_filename.parquet"`)
* Python scripts are both importable as a module (same directory....use `import filename`) and runnable (in terminal we can type `python some_script.py`.
   * You can import from other directories, but you have to add bells and whistles to it. We'll cover it in the future.
* Distinction between importing a function to use elsewhere vs multiple functions you want to run together to generate raw -> cleaned data.
   * Compare this [notebook](https://github.com/cal-itp/data-analyses/blob/main/la_metro_demo/02-select-division.ipynb) which has an import line `import A3_assemble_data as assemble_data` with [this script](https://github.com/cal-itp/data-analyses/blob/main/facilities_services/A9_create_processed_data.py).
   * `import A3_assemble_data as assemble_data` means that I am importing a function from the script `A3_assemble_data`, but I'd like to import it as a module. I am referring to this module as `assemble_data`. We do abbreviations all the time: `import pandas as pd`
   * `A3_assemble_data` has no `if __name__ == "__main__"` section, so nothing happens when you run `python A3_assemble_data.py` in the terminal. 
   * Every time I run `import A3_assemble_data as assemble_data`, every function is "run" and returns the df at the end. If the function is not "run", how can it construct a df?
   * There are times when it's not desirable to keep running a function to generate data. If it's a small df, it's not a big deal, but if it takes 10 min to run, you definitely are not wanting to run that 10 min function every time you are in a notebook. You'd save out your result in the middle and then call in the already-saved, cleaned df to start. For a small df with minimal transformations, you can keep your functions operating on the raw df and calling it in the notebook as a cleaned df (sometimes we do it to minimize sprawl in our bucket).
   * This [script](https://github.com/cal-itp/data-analyses/blob/main/facilities_services/A9_create_processed_data.py) has a `if __name__ == "__main__"` section.  When I run `python A9_create_processed_data.py`, it will run through those steps below and save out a parquet file. Those "steps" include reading in raw data, cleaning it with functions (some defined in the same script, some are imported from `utils`) and saving it out.

Action items:
1. Add `if __name__ == "__main__":` section to cleaner scripts. I think the amount of data cleaning is not minimal, there's quite a lot of replacing of values, so you'd want to save it out at the end.
2. Choose descriptive names for functions.  
   * Ex: `dgs_data_cleaner`. One step is cleaning up the values. You can define that as `def clean_dgs_columns(df):`, where it takes some unspecified df. Then a second function for aggregation: `def aggregate_by_agency(df)`. 
   * A general rule of thumb: use different functions to do different things, especially when you're changing the rows dramatically (aggregation). 
   * The first function was data cleaning, like, cleaning up nulls, wrangling strings into floats, dropping missing rows, etc etc. 
   * The second function is taking that cleaned up function and aggregating...which by nature reduces the rows quite dramatically. I would separate it into a different function. Later, if you're investigating why one agency is not in a chart, you can look whether it was dropped because of missing values in the first step or because in the aggregation, it was left with zeros, and thus not displayed in the chart. 
3. Proper indentation. Things inside parentheses are 1 indent in.
```
df = pd.merge(
    df1, 
    df2, 
    on = "my_column"
)
```

Issue #936, opened 2023-10-26
Research Request - RT vs Schedule Trip Level Metrics (using vp_usable)


amandaha8:
Complete the below when receiving a research request, and continue to add to this issue as you receive additional details and produce deliverables. Be sure to also add the appropriate project-level label to this issue (eg gtfs-rt, DLA).

## Research Question
**Single sentence** description:  Add metrics to the all the vp_usable files going back to March. 
* Relates to Epic #824 
  
**Detailed** description: A lot of metrics work has already been done, but not necessarily starting from the same dataset which causes discrepancies. These metrics will now be created using the same vp_usable file. 

* [Work in this folder.](https://github.com/cal-itp/data-analyses/tree/main/rt_scheduled_v_ran)
* Use the `trip_instance_key` column to join schedule and RT trips.
* [Use pd.Grouper](https://pandas.pydata.org/docs/reference/indexing.html)
* Use groupby(observed = True)

## How will this research be used?
Determine the quality of the GTFS data (on the trip level) an operator produces. 

## Stakeholders & End-Users
Who is asking the question, and who is the audience?

## Metrics
* Spatial accuracy: do the vehicle positions fall within the recorded route shape we have? If they don't (to a certain degree), delete these trips out. [Script](https://github.com/cal-itp/data-analyses/blob/main/rt_scheduled_v_ran/scripts/vp_spatial_accuracy.py)
* How many minutes a trip took and the average speeds? [Script 1](https://github.com/cal-itp/data-analyses/blob/main/rt_segment_speeds/scripts/C2_triangulate_vp.py) [Script 2](https://github.com/cal-itp/data-analyses/blob/main/rt_segment_speeds/scripts/C3_trip_route_speed.py) 
* Out of the total minutes a trip took, how many of those minutes had more than 2 vehicle positions? [Code](https://github.com/cal-itp/data-analyses/blob/main/rt_predictions/01_update_completeness.ipynb)

## Data sources
* **Cal-ITP data sources**:
*  [vp_usable file. ](https://console.cloud.google.com/storage/browser/calitp-analytics-data/data-analyses/rt_segment_speeds?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&project=cal-itp-data-infra&prefix=&forceOnObjectsSortingFiltering=false)
* **External data sources**:

* **Remaining data source questions**:

## Deliverables
Scripts.

## Timeline of deliverables
Estimated completion date

PR #1033, opened 2024-02-20
RT vs Schedule 

amandaha8:
- Added in `schedule gtfs key` `route_id` `dir_id` `time_of_day` and `offpeak_onpeak` columns to `routes` and `trips`.
- Separated workstream between `routes` and `trips`.  Metrics are always generated on a `trip` level. `Routes` is calculated with a weighted average and any metrics from `trip` are deleted before added back in. 
- Saved outputs to the correct folder in GCS's `rt_v_schedule` folders. 
- New function that adds in `time_of_day` in `segment_speed_utils` based on `vp` positions instead of `schedule`.  If the trip is only found in `vp_usable` and not in both `vp_usable` and `schedule`, we can determine now the `time_of_day` based on `vp`.


rt_scheduled_v_ran/scripts/rt_v_scheduled_routes.py
```diff
@@ -0,0 +1,108 @@
...
+    sum_cols = [
+        "total_min_w_gtfs",
+        "rt_service_min",
+        "total_pings",
+        "service_minutes",
+        "total_vp",
+        "vp_in_shape",
+    ]
+
+    count_cols = ["trip_instance_key"]
```
amandaha8, line 79: @tiffanychu90 Can you tell me if I added too many columns in the group_cols list?

tiffanychu90: I think you just need `group_cols = ["schedule_gtfs_dataset_key", "route_id", "direction_id"]`. 
I think `peak_offpeak` and `time_of_day` are going to force the granularity to be more fine than we want. It'll leave rows like `route A - direction 0- AM_peak - peak`, `route A-direction 0 - Midday-offpeak`. which is too granular. We want `route A - direction 0 - peak`, `route A - direction 0 - offpeak`, `route A - direction 0 - all_day`

tiffanychu90:
* #936
* #966

PR #1037, opened 2024-02-23
Refactor dgs_data_cleaner script

csuyat-dot:
Refactored `dgs_data_cleaner` script following comments/suggestions. Addressed the following:

- Imported variables from other scripts
- Wrapped a lot of the individual steps from the notebook into functions. 
- `clean_dgs_columns` now performs: reading in data, merging sources, snakecase function from shared utils, update datatypes for columns, created new columns based on other functions.
- `new_prop_finder` and `new_bus_size_finder` functions now use multiple lists to check against, then returns a normalized values
- Reworked `agg_by_agency` and `agg_by_agency_w_options` functions. Since the DGS data contain rows for different items (bus, equipment, options), it was necessary to have multiple outputs. 1 table that aggregates the cost of buses ONLY, and another table that aggregates the cost of buses with options. Different agencies can configure buses with different options, which causes the total cost to vary. Having 2 outputs ensure we are prepared if the specifications of the analysis changes.

Next, will refactor the remaining scripts: `tircp_data_cleaner` and `fta_data_cleaner`

### DIFF
--- a/bus_procurement_cost/dgs_data_cleaner.py
+++ b/bus_procurement_cost/dgs_data_cleaner.py
@@ -151,7 +151,7 @@ def agg_by_agency(df:pd.DataFrame) ->pd.DataFrame:
 if __name__ == "__main__":
 
     # initial df
-    dgs_17bc = clean_dgs()
+    dgs_17bc = clean_dgs_columns()
     
     # assigning new columns
     df2 = dgs_17bc.assign(
--- a/bus_procurement_cost/dgs_data_cleaner.py
+++ b/bus_procurement_cost/dgs_data_cleaner.py
@@ -4,168 +4,351 @@ import shared_utils
 from calitp_data_analysis.sql import to_snakecase
 from fta_data_cleaner import gcs_path
 
-# params
-file_17c = "17c compiled-Proterra Compiled Contract Usage Report .xlsx"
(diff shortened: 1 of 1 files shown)


tiffanychu90:
Functions are a lot cleaner and more usable! Good job!
* I would start tweaking some of the categories because `battery electric`, `electric`, `fuel cell electric`, `hybrid electric` all share the word `electric`. I would watch out, because this means it's fairly sensitive to the order of the if/elif/else conditions.
   * `if vs elif`. `if` goes through and checks every condition. `elif` checks it if a previous condition isn't met. 
   * Ex: A battery electric bus checks whether it's `battery electric` and it's categorized as `BEB`. Then later, fuel cell and hybrid cells don't match, so it doesn't get overwritten. For `battery electric`, that's where you want it to end. But if it finds `electric` in the description, it's possible that it can be recategorized as `electric (not specified)`. 
* You might want to include an extra `not condition` like [here](https://github.com/cal-itp/data-analyses/blob/main/dla/dla_utils/clean_data.py#L349-L351). The logic would be: if you have the word `battery` and not `diesel` and not `hybrid` and not `fuel cell`, you can be categorized as `battery electric`. The last condition would be an `else`, if you're still not categorized as anything, return `electric (not specified)`.

PR #1040, opened 2024-02-29
Aggregating vp_usable trips to routes

amandaha8:
- Updated work per Tiffany's feedback. 
- Feedback #1: updated the grouping columns for aggregating trips up to routes so it would be less granular. 
- Feedback #2: adjusted the routes aggregation function so the routes are rolled up by peak/offpeak. Added the `time_period` column with values that match with other datasets such as  [this](https://github.com/cal-itp/data-analyses/blob/main/rt_segment_speeds/scripts/config.yml#L9) and [this](https://github.com/cal-itp/data-analyses/blob/main/rt_segment_speeds/scripts/config.yml#L6C5-L6C31).
- Double checked that the `routes_df` merged nicely with outputs [here](https://github.com/cal-itp/data-analyses/blob/main/gtfs_digest/merge_data.py#L121-L122)
- Begin troubleshooting why San Francisco Muni is totally missing?


tiffanychu90:
* #966

PR #1044, opened 2024-03-06
bus cost scripts refactor

csuyat-dot:
continue work on...
- #897 

refactored `fta_data_cleaner` script by removing old cleaning steps and brining in cleaning functions from other scripts. specifically, `new_bus_size_finder`, `new_prop_finder` and `col_row_updater` from `dgs_data_cleaner` and `tircp_data_cleaner`.

`from fta_data_cleaner import gcs_path` was causing circular import problems in the other scripts. resolved by moving  the `from... import...` to within the functions.

refactored `cost_per_bus_cleaner`. reads in dgs, fta and tircp data and merges them all together. 

created a new script called `cpb_analysis_scripts` that holds all the functions to be used in the `cost_per_bus_analysis` notebook, 

finally, refactored the notebook by removing all the old functions and reorganizing the variables into similar categories.


tiffanychu90:
A couple of comments:
* `cpb_analysis_scripts.py` - combine several steps into 1 function with a more readable name (cpb isn't obvious until the docstring). the cost per bus, the z-score, and removing of outliers can all be done together. Is there a reason to separate them? I can see all 3 steps needed across each df you're bringing in.
   * `df1 = cost_per_bus(df1) , df2 = cost_per_bus(df2)`
* I think this [line ](https://github.com/cal-itp/data-analyses/blob/main/bus_procurement_cost/dgs_data_cleaner.py#L148C42-L148C59)is written wrong, specifically `.replace("-" "", " ")`. Check the docs to see what you want to do here.
* [This](https://github.com/cal-itp/data-analyses/blob/main/bus_procurement_cost/dgs_data_cleaner.py#L80-L86) is redundant.
   * If you are looking for this substring `battery electric`, it would find `battery electric bus` and `battery electric buses`...so you don't need to look for `battery electric bus` and `battery electric buses` again. To remove extra space, you want `.str.replace('[two spaces]', '[one space]')`

PR #1050, opened 2024-03-19
Mar metadata

tiffanychu90:
## open data
* Follow-up to #1048 
* Overwrite layers again in enterprise gdb just to be sure it's not us missing the metadata (metadata appears to be there)
* Continue to separate the metadata XMLs with sub-folder and undo last PR
* Given that open data portal is having issues with layers stored in .sde (which we use), we'll will add HQ transit stops for March 2024 to public GCS as a zipped shapefile (many housing user requests!)
```
I am using the HQrail(edit)@sv03tmcsqlprd1.sde database. However, I just discovered an issues with this 
database that just started. Today, it was working last week and when I was trying to update the week before. 
Also, yours is not the only one, but it is not occurring for all of the sde databases on this server.
```

## segment speeds / rt_stop_times
* Technically, a trip summary speed should be calculated from first and last stop of a trip
* We'll do that and use the `rt_stop_times` speeds file created to do summary speeds
* Leave the segment speed averaging as is.
   * `average_speeds` script, last commit hash: 7e7a0512099eab5494bc8798e2abf88f1c166f3f 
   * oops, `git mv` not done successfully and now history is broken


tiffanychu90:
Close - separate out topics within PR

PR #1068, opened 2024-03-29
Bus cost makefile and notebook update

csuyat-dot:
continue work on issue:
- #897 

Updated `bus_cost_analysis.ipynb` to show total number of projects and number of bus only projects for each dataset. Also fixed the number of buses variable.

Updated `Makefile` to include `dgs_data_cleaner` and nbcovert commands to run notebook and export as HTML to `bus_cost_analysis.html`

Updated `README.md` file to explain scripts in makefile.


tiffanychu90:
* Within your functions, if you're returning `df1`, which is `df` + some transformations / filtering, etc, you don't need a line right before that is `df1 = df.copy()`. Using a function that takes `df` as an input means you want to transform it and return something else, there's no need to make a copy of it...`df` already exists outside of the function. You definitely expect `df` and `df1` to be different! You can remove these extraneous references.
* I like the use of a notebook turned into a readable report, constructing your text with injected variables, using `nbconvert` to turn it into an HTML page!
* Make sure the final written narrative has gone through spelling and grammar checks
* Refactor overall “pipeline” / concepts / coherent and concise thinking using what we outlined in the Google Docs and the `outline.md` I checked into the folder

Issue #1059, opened 2024-03-26
Research Request - GTFS Digest Metrics Shakedown


tiffanychu90:
Complete the below when receiving a research request, and continue to add to this issue as you receive additional details and produce deliverables. Be sure to also add the appropriate project-level label to this issue (eg gtfs-rt, DLA).

## Research Question
**Single sentence** description: Draft GTFS digest is up, now decide which metrics are worth charts (time-series), which are more easily captured by tables (perhaps mostly static with little changes) 
* #989 

#### Grain
* route-direction
* in digest, do not use `schedule_gtfs_dataset_key` or `name (schedule)`, but use `organization_source_record_id` and `organization_name`. 
* organize by district
* meaningful route name displayed, and charts use this standardized `route_name`, not `route_id`:
   * put through script to standardize / parse route_id over time so we can identify the same route over time (route_id2). 
   * further clean up by operator so that combo `route_short_name` and `route_long_name` does not have redundancy (`1 Route 1` -> `Route 1`)

#### Operator Stats ("digest/operator_profiles" and "digest/operator_routes")
- [x] monthly scheduled service hours by day of week (`day_type`, `time_of_day`)
- [x] number of routes, service area (length in miles, only count each route-direction once)
- [x] number of stops served, total stop arrivals, arrivals per stop 
- [x] route typology breakdown

#### Schedule ("digest/operator_schedule_rt_category"?) 
- [x] avg scheduled service minutes
- [x] avg stop distance (meters) -> change to per mile? 
- [x] all_day / peak / offpeak scheduled n_trips
- [x] all_day /peak / offpeak scheduled frequency 
- [x] monthly scheduled service hours by day of week (`day_type`) and `time_of_day`

#### Speeds ("digest/segment_speeds") 
- [x] summary speeds for all_day /peak / offpeak

#### RT vs Schedule
- [x] trips breakdown by `sched_vp_category` (10 trips schedule only, 100 trips schedule_and_vp, 5 trips vp only)
   * `.drop_duplicates()` but keep some dates where this distribution changes
- [ ] <s>route breakdown by `sched_vp_category` ("digest/schedule_vp_metrics") </s> don't do this, because routes can have different numbers of `n_scheduled_trips`, `n_vp_trips`, and we can assign incorrectly.

#### Schedule
- [x] monthly scheduled service hours by day of week (`day_type`) and `time_of_day`
- [x] number of trips by `sched_vp_category`?

#### Speeds
[...]

PR #1076, opened 2024-04-12
GTFS Exploration

amandaha8:
- Updated the existing charts using the right filters and data sources.
- Ran all the operators with the test notebook to see if any bugs arose -> discovered that concatenating all those charts will cause Altair to error out because this can total to a lot of data depending on the operator. In this case, it was AC Transit that caused this error. This was fixed by turning off Altair's warning but could pose problems in the future since this is a notebook with a lot of charts.
- Took a look at the other data sources that will feed into this portfolio.
- Sketched out a gameplan and to-do based on the first run of all the operators.


tiffanychu90:
* #1059

PR #1079, opened 2024-04-18
GTFS Digest Portfolio 

amandaha8:
- First version of section 1, which is an overview of the operator using the most recent month's data.
- Reworked section 2 to incorporate feedback. I changed the color scheme for some of the graphs, manually set the axis, updated graphs that were confusing, among other changes.
- Added new rows to the yaml file that acts as the source of truth for renaming dataset columns to something "public facing" and where graph titles/subtitles are pulled from. 
- Reran parameterized nb.


tiffanychu90:
* #1059

PR #1085, opened 2024-04-26
GTFS Portfolio

amandaha8:
- I noticed there were multiple organization name-name-schedule key combinations that were duplicative. As an example: the organization_name "City of Eureka" could correspond with "Eureka Schedule" or "Humboldt Schedule." This caused errors when generating the parameterized pages. I fixed this by finding the rows with the most recent `service_dates` and using an inner merge to ensure the organizations included had rows across all the datasets in the yml generation script.
- Updated the `readable.yml` to rename the dataframes harmoniously and act as a "source of truth" for all the chart titles/subtitles.
- Made aesthetic tweaks such as commas for large numbers, changing the y axis limit, making the color scheme prettier.
- Delineated Section 2 between charts that showcase the quality of a rider's experience on that route vs the quality of the data generated.
- Began more work on Section 1's Service Hour chart: differentiate between daily versus total service hours for an operator based on feedback. 
- Unfortunately hit a roadblock, the charts in section 2 now do not generate properly after applying the `readable.yml` to it and streamlining stuff. A Javascript error message pops up the first time a cell is ran, but the second time, the charts are generated. One roundabout way is running the jupyter cell twice programmatically but this is not ideal since it doesn't account for the underlying issue and the error still shows up.


tiffanychu90:
* #1059

Issue #1060, opened 2024-03-26
Research Request - GTFS Digest Data Cleaning


tiffanychu90:
Complete the below when receiving a research request, and continue to add to this issue as you receive additional details and produce deliverables. Be sure to also add the appropriate project-level label to this issue (eg gtfs-rt, DLA).

## Research Question
**Single sentence** description:  There are several columns that do not exist yet, and other columns that need more data processing for GTFS digest. 
* Data cleaning supporting #1059
  
### GTFS Digest pipeline
- [x] make sure we move to `organization_source_record_id` and `organization_name` after we run `merge_data.py`. Make sure column is present to use in notebook.
- [x] organize by district -> update `gtfs_digest/deploy_portfolio_yaml.py`
- [x] meaningful route name displayed, and charts use this standardized `route_name`, not `route_id`:
   * further clean up by operator so that combo `route_short_name` and `route_long_name` does not have redundancy (`1 Route 1` -> `Route 1`)

### Schedule Pipeline
- [x] Heuristic to assign each route one of the NACTO route typologies (`gtfs_funnel/route_typologies.py`):
   * [transit route types](https://nacto.org/publication/transit-street-design-guide/introduction/service-context/transit-route-types/) and [transit frequency](https://nacto.org/publication/transit-street-design-guide/introduction/service-context/transit-frequency-volume/)
- operator stats aggregation
   - [x] monthly scheduled service hours by day of week (`day_type` use readable name, `time_of_day`)
   - [x] number of routes, service area (length in miles, only count each route-direction once)
   - [x] number of stops served, total stop arrivals, arrivals per stop 
   - [x] route typology breakdown

#### Schedule
- [x] avg stop distance (meters) -> change to per mile in `gtfs_funnel/schedule_stats_by_route_direction.py`

#### RT vs Schedule
- [x] do some of the pre-aggregations for trips and route breakdown by `sched_vp_category` (10 trips schedule only, 100 trips schedule_and_vp, 5 trips vp only)
- [x] % RT journey with 1+/2+ vp (are the correct averages getting generated? these should be straight averages across trips?)
- [x] % schedule journey with 1+/2+ vp (are the correct averages getting generated? these should be straight averages across trips?)
- [x] comparison of scheduled to RT trip time (aggregated to route-direction)
   * breakdown n_trips early (5+ min early) / on-time within 5 min early or late / late (5+ min late)

## Deliverables
[...]

Issue #989, opened 2024-01-08
Epic - GTFS Digest (exploratory portfolio)


tiffanychu90:
After receiving a research request, use this template to plan and track your work. Be sure to also add the appropriate project-level label to this issue (eg gtfs-rt, DLA).

### Summary
* Track the portfolio work displaying everything we can display with GTFS schedule, speeds, RT vs schedule across all our 4 major workstreams within GTFS (schedule, speeds, RT vs schedule, predictions)
* Outline all the differences in columns so we can land on a standardized way to approach this
   * Make sure every dataset has similar columns so we can merge in our time-series datasets 
   * Keep track of the tables we have created  / aggregated and use this to inform any new tables we are creating

### Speed Tables

**GCS** 
* `SEGMENT_GCS/rollup_singleday/` 
* `SEGMENT_GCS/rollup_multiday/`
* [Makefile](https://github.com/cal-itp/data-analyses/blob/main/rt_segment_speeds/scripts/Makefile) and [data catalog](https://github.com/cal-itp/data-analyses/blob/main/_shared_utils/shared_utils/gtfs_analytics_data.yml)

| **dataset** | **grain** | **geometry** | **time_category** | **single_multi** | **operator_ids** | **timeseries_postprocessing** |
|---|---|---|---|---|---|---|
| shape-stop_segments | (shape_array_key)  shape_id stop_sequence | segment_geom (select one trip per shape) | peak,  off peak, all-day | single <br>multi | (gtfs_dataset_key) org_source_record_id, org_name | cannot be used for multiple  months, as shape_id will appear distinct even in a  6 month period |
| route_direction- stop_segments | route_id, direction_id,  stop_pair | segment_geom | peak,  off peak, all-day | single | (gtfs_dataset_key) org_source_record_id, org_name | put through script to  standardize / parse route_id over time so we can identify the same route over time  (route_id2) |
| route_direction- stop_segments | route_id, direction_id, stop_pair | segment_geom  | peak, offpeak, all-day weekday, weekend, month, year | multi | (gtfs_dataset_key) org_source_record_id, org_name | put through script to  standardize / parse route_id over time so we can identify the same route over time  (route_id2) |
| trip avg speed | trip_instance_key | tabular | has time_of_day column,  but is disaggregated | single | gtfs_dataset_key |  | not used beyond single day |
[...]

PR #1093, opened 2024-05-02
GTFS Digest

amandaha8:
- Imported NTD data from our warehouse to incorporate some basic information about the operators like what type of organization it is, its headquarters, etc.
- Updated `README.md` with the GCS download link and link to methodologies.
- Made tweaks such as ensuring the legend did not run off, resizing charts, adding more headers, changing width of bars, etc.


tiffanychu90:
* #989 
* finish up #1059 and add `README` for #1060

Issue #1094, opened 2024-05-02
Research Request - SACOG Questions regarding NTD Monthly Ridership by RTPA


csuyat-dot:
Complete the below when receiving a research request, and continue to add to this issue as you receive additional details and produce deliverables. Be sure to also add the appropriate project-level label to this issue (eg gtfs-rt, DLA).

## Research Question
**Single sentence** description:  

- Respond to questions via SACOG email regarding specific visuals on [NTD Monthly Ridership webpage on Cal-ITP portfolio](https://ntd-monthly-ridership--cal-itp-data-analyses.netlify.app/readme)

  
**Detailed** description:
- Specific questions are:
1. "El Dorado County Transit only date through June of 2019 is shown, but data are available through June of 2022. ... Is there a way to add the information from the small non-NTD monthly reporters to what is presented if we assisted in the collection of those data?"
2. in the section on change in unlinked passenger trips year-over-year it appears that Yolo Transportation District is no included, but should be since they are a monthly reporter on UPT.
3. Do you know if the demand response category is lumping in microtransit/on demand service with demand response service? It may be interesting to show the recent changes/expansions in microtransit service availability

## How will this research be used?
What will be achieved or informed? What is the intended impact?

## Stakeholders & End-Users
Who is asking the question, and who is the audience?
- Barbara VaughanBechtold. Capitol Valley Regional SAFE Program Mgr./Transportation Planner at SACOG

## Metrics
* any specific calculations
* you think will be needed

## Data sources
* **Cal-ITP data sources**:
*  https://ntd-monthly-ridership--cal-itp-data-analyses.netlify.app/readme

* **External data sources**:
* https://www.transit.dot.gov/ntd/data-product/monthly-module-adjusted-data-release

* **Remaining data source questions**:

## Deliverables
Notebook, dashboard, narrative, presentation details.
- email and/or meeting with Barbara

## Timeline of deliverables
Estimated completion date

csuyat-dot:
related issue regarding report tweaks
- #990

csuyat-dot:
[...]

PR #1117, opened 2024-05-17
Monthly Ridership by RTPA chart adjustment

csuyat-dot:
related PR 
- #1094 

Initially, charts would not show negative changes in UPT values. We should show negative changes and net zero changes in UPT, but not if UPT was zero for consecutive years.

developed function that removes rows with zero UPT data for the current and prior year. also, removed the filtering condition for `make_bar_chart` function so that negative and valid zero UPT values are shown.


csuyat-dot:
tested new `remove_zero_upt_rows` function and updated chart criteria against initial function/chart. summary of impact to change in UPT charts are below:

- 24 total RTPA pages
- 15 RTPAs pages stayed the same (no changes)
- 9 RTPA pages had changes (4 new agencies appeared, 10 new negative bar chart values appeared)

PR #1124, opened 2024-05-30
GTFS Digest Portfolio

amandaha8:
- Fixed the direction column so all 0/1 have corresponding cardinal directions.
- Reran all the operators.


tiffanychu90:
## folding into analytics pipeline 
* consult the diagrams available in `gtfs_funnel` and decide what grain [route cardinal direction / but any dataset you want to add] you **need** for GTFS digest, and identify what grain it **is** in your most upstream df.
* the diagrams show that after a single date is downloaded for schedule (4 tables) and RT (1 vp table), they are processed topically.
   * schedule data - small enough so it's done within `gtfs_funnel`, as some of these are needed in segment speeds
   * segment speeds (vp) - a separate topic because it's it's trying to handle many types of segments, and its intermediate grain is trip-some_segment_definition. After we select a kind of segment, it's fairly geospatially intensive to scan through each vp row to get what we want (find the specific position that is nearest to a stop, and then find the corresponding timestamp). 
   * rt vs schedule (vp and schedule) - even though this topic combines the above 2, it's processed as a separate workstream because it's intermediate grain is trip-minute or trip, and not about trip-segment.
* decide where your step fits into the above workstreams
   * if you can add a column to an existing df, then do it!
   * if you can't, because your upstream df is a different grain than what it **needs** to be, then add a separate function and aggregate to the grain you need it to be
   * hint: schedule data is aggregated to route-dir almost immediately (upstream table is `trips`, and in gtfs digest it is `route-direction`. cardinal route direction's upstream table is [fill in the blank], and in gtfs digest it is `route-direction`. 
   * hint 2: since gtfs digest handles a single date each month, you will want to work 1 date for cardinal direction in, and let that go through the established processing (aka, `clean_route_naming`, `time_series_utils.concatenate_across_dates` and not do it yourself.  
* let's discuss after you've though about all of these steps and work through where there's confusion!
* make sure to attach relevant GH issues to this PR!

Issue #1027, opened 2024-02-15
Epic - Monthly NTD update


tiffanychu90:
After receiving a research request, use this template to plan and track your work. Be sure to also add the appropriate project-level label to this issue (eg gtfs-rt, DLA).

### Summary
* Track all the monthly for NTD ridership PRs in one place.
   * Every month we pull the NTD data that is 2 months prior. In October 2023, we're pulling August 2023's data.  
* Do this until we have Airflow DAGs. 

## PRs

**monthly report publishing**
- Oct 2023: #978 (initial task)
- Nov 2023: #1008 
- Dec 2023: #1030 
- Jan 2024: #1049 
- Feb 2024: #1090 
- Mar 2024: #1114 
- April 2024: #1136 
- May 2024: #1170 
- June 2024: #1205
- July 2024: #1222 
- Sept 2024: #1287
- Oct 2024: #1320 
- Nov 2024: #1346 
- Dec 2024: #1393 
- Jan 2025: #1418
- Feb 2025: #1454 
- March 2025: #1495 
- April 2025: #1506 
- May 2025: #1564 
- June 2025: #1631 
- July 2025: #1709 (updated to new warehouse table)
- Aug 2025: #1736 
- Sept 2025: #1790 
- Oct 2025: #1815 
- Nov 2025: #1871 
- Dec 2025: #1940 
- Jan 2026: #1971 
- Feb 2026: #2010 
- March 2026: #2047 
- April 2026: #2089 

**bug fixes and features**
- #990 
- #1165
- #1117 
- #1190 
- #1197 
- #1260 
- #1328 
- #1516 
- #1524
- #1525 
- #1534 
- #1612 
- #1657

csuyat-dot:
Discussed with @tiffanychu90 . The Monthly and Annual NTD ridership reports can now be found in:
- https://github.com/cal-itp/ntd-snapshot

Updates to the report files are expected to be minimal from now on (maybe the report .yml file may change if RTPAs are removed or added). No longer need to update configuration .yml with the current month, no longer need to check in portfolio site files. 

As of 9/14, just published the June 2026 report and data.

closing epic.

PR #1136, opened 2024-06-07
April 2024 Monthly Ridership NTD publish

csuyat-dot:
Received initial error stating 2 rows did note merge between NTD data and NTD ID crosswalk.
investigation found that Kern Regional Transit appeared in the NTD data but not in the crosswalk.
Added new row to crosswalk with Kern Regional Transit data.
Re-ran the make ntd_monthly_export script without any errors.
Reviewing the draft portfolio site confirms: 
1. Kern Regional Transit appears under the Kern Council of Governments RTPA page. 
2. New formula from previous fix (showing negative numbers) is working as intended. see #1117 

Also confirming the csv .zip files produced correctly in the public GCS bucket & other related data produced in the private GCS


tiffanychu90:
* #1027

Issue #1135, opened 2024-06-07
Research Request - add cardinal direction into GTFS analytics pipeline


tiffanychu90:
Complete the below when receiving a research request, and continue to add to this issue as you receive additional details and produce deliverables. Be sure to also add the appropriate project-level label to this issue (eg gtfs-rt, DLA).

## Research Question
**Single sentence** description:  Now that `cardinal_direction` as a readable description can be used instead of `direction_id=0/1` in GTFS digest, work on adding this into the pipeline. 
* Analytics pipeline supporting #1101
  
**Detailed** description: 
- see notes in #1124
- [ ] identify scripts that need this new column added
- [ ] test scripts from upstream to downstream on smaller set of dates
- [ ] if digest looks good, backfill remaining dates
- [ ] if digest needs to create new datasets in a different grain, do so
- [ ] deploy portfolio, so pipeline and product are seamless
- [ ] publish the datasets used in GTFS digest to the public GCS bucket (including any new datasets created) 
- [ ] supporting docs / diagrams / descriptions, etc

## Deliverables
Take the functions outlined in a notebook and slot them into the correct stages of transformations backing GTFS digest.

PR #1145, opened 2024-06-13
Cardinal Direction 

amandaha8:
- In response to #1135 
-  Integrated cardinal direction into all the relevant  files after testing on a smaller set of dates. Reran all the impacted files until April 2024, since May and June are still under construction.
- Deleted out old work that adds cardinal direction at the end of the `digest/schedule_vp_metrics`. Double checked and made sure charts still popped up. 
- Updated the `requirements.txt` in `shared_utils` since I'm always upgrading to the latest `Altair` and `Altair Transform` versions.


gtfs_digest/merge_data.py
tiffanychu90, line 1: Can you switch `route_time_with_cardinal_dir_cols` to `route_time_cols +  ["route_primary_direction"]`. If the route groupings change, you'll want to check just 1 reference, rather than multiple places. If `sort_cols_with_cardinal_dir_cols` is just referenced once, just use `sort_cols + ["route_primary_direction"]`.

amandaha8: I think I may have misunderstood you yesterday, I made a distinction between `route_Time_cols` and `route_time_with_cardinal_dir_cols` because I thought `concatenate_speeds_by_route_direction` and `concatenate_rt_vs_schedule_by_route_direction` don't have cardinal direction info?

gtfs_funnel/schedule_stats_by_route_direction.py
tiffanychu90, line 1: The functions for counting stop directions can benefit from some streamlining. I actually think you can use a standalone function and get it to the route-dir aggregation and merge it in within `schedule_metrics_by_route_direction()` Sticking it into `assembled_scheduled_trip_metrics()` means you want it to be at the trip grain, but the cardinal direction is not even calculated at the trip grain, hence, moving it into its own function / using it directly in `schedule_metrics_by_route_direction()` might be more straightforward. 

Here's some pseudo-code: 
```
# make use of columns and filters kwargs
def cardinal_direction_for_route_direction():
   # i think the docstring here can be clear on the method, and the entire aggregation now can 
   # sidestep the entire trip grain (the trip table is only useful for getting route_id and direction_id)
   df = gpd.read_parquet(STOP_TIMES_FILE, 
      columns = keep only the columns you need + ones to merge for trips, 
      filters =filter out unknowns )

   trips = helpers.import_scheduled_trips(columns = )

   stop_times_with_trip = pd.merge(df, trips, on = "trip_instance_key")

   cardinal_direction = (stop_times_with_trip.groupby(route_dir_cols)
      .agg({"stop_primary_direction": "count"})
      .sort_values(descending)
      .drop_duplicates(subset=route_dir_cols_)

   return cardinal direction
```

gtfs_funnel/schedule_stats_by_route_direction.py
```diff
@@ -40,28 +128,44 @@ def assemble_scheduled_trip_metrics(
...
         how = "left"
     ).merge(
-        trips_to_route,
+        trips_to_route[trips_to_route_cols_subset],
         on = "trip_instance_key",
         how = "inner"
     )
     
-    return df2
+    median_stop_meters_df.direction_id = median_stop_meters_df.direction_id.fillna(0)
```
tiffanychu90, line 149: What if there are nulls in `cardinal_direction_df`? would this NaN filled in prevent some merges?

gtfs_funnel/schedule_stats_by_route_direction.py
```diff
@@ -40,28 +128,44 @@ def assemble_scheduled_trip_metrics(
...
+    
+    return m1
     
     
 def schedule_metrics_by_route_direction(
     df: pd.DataFrame,
     analysis_date: str,
-    group_cols: list
+    group_cols: list,
+    merge_cols:list
```
tiffanychu90, line 168: Once your cardinal direction function is added, you no longer need `merge_cols` as an arg.

gtfs_funnel/schedule_stats_by_route_direction.py
amandaha8: I filled out any nan values in `direction_id` as 0 in `cardinal_direction_df`, so I think it should be ok?

PR #1154, opened 2024-06-21
Adds ntd ridership numbers and impact analysis

vevetron:
- Added a query that pulls ridership numbers
- Adjusts LA metro bus and rail to get close to matching their distributions
- Removes duplicate ntd entries
- Creates graph and tables with estimated ridership impacts.


holiday_service_research/holiday_research.ipynb
```diff
@@ -230,6 +245,177 @@
...
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "upt = query_sql(f\"\"\"\n",
+    "SELECT year, ntd_id as ntd_id_2022, agency_name, reporter_type, time_period, sum(unlinked_passenger_trips__upt_) as sum_unlinked_passenger_trips_upt  \n",
+    "FROM `cal-itp-data-infra.mart_ntd.dim_annual_ntd_agency_service` \n",
+    "where ntd_id in ({ntd_ids_for_query})\n",
+    "and time_period = 'Annual Total'\n",
+    "and year = 2022\n",
```
evansiroky, line 400: Does this query need to be modified to filter out inactive and demand-response ridership?

vevetron: It doesn't need to filter out inactive ridership.

I lumped in the demand-response ridership with the regular fixed-route ridership.  I assumed it's more likely the holiday service would affect demand-response as well so any impact of bad schedules would affect both regular fixed-route and demand-responose the same.

evansiroky: ah, ok. Seems like a reasonable assumption.

evansiroky (approved):
Looks great. Thanks for all of this work!

PR #1156, opened 2024-06-25
Cardinal Direction 

amandaha8:
* In response to #1135.
* I made the changes detailed in #1145 
* Updated lists of columns for grouping and merging so there is less repetitiveness. 
* Separated the function that finds the cardinal_direction out of `assembled_scheduled_trip_metrics()` and into its own function, since it is on the route grain.
* Merged the results from the new `cardinal_direction_for_route_direction()` directly into `schedule_metrics_by_route_direction()`.
* Experimented with filling in the `nans` for the `direction_id` column in `cardinal_direction_for_route_direction()` and `assemble_scheduled_trip_metrics`: it only seems to add more rows for `assemble_scheduled_trip_metrics` when the dataframe gets grouped together.
* Deleted out group and merge_cols.

### DIFF
--- a/gtfs_funnel/schedule_stats_by_route_direction.py
+++ b/gtfs_funnel/schedule_stats_by_route_direction.py
@@ -12,81 +12,6 @@ from segment_speed_utils import helpers, gtfs_schedule_wrangling
 from shared_utils.rt_utils import METERS_PER_MILE
 from update_vars import GTFS_DATA_DICT, SCHED_GCS, RT_SCHED_GCS
 
-def find_most_common_dir(
-    stop_times_gdf: gpd.GeoDataFrame,
-    trips_to_route_df: pd.DataFrame,
-) -> pd.DataFrame:
-    """
-    Instead of only having direction_id as 0/1, add back the
-    cardinal directions. 
-    """
-    stop_times_col = [
-        "feed_key",
-        "stop_id",
-        "stop_sequence",
-        "schedule_gtfs_dataset_key",
-        "trip_instance_key",
-        "shape_array_key",
-        "stop_name",
-        "prior_stop_sequence",
-        "subseq_stop_sequence",
-        "stop_pair",
-        "stop_pair_name",
-        "stop_primary_direction",
-        "stop_meters",
-    ]
-    
-    # Keep only a subset
-    stop_times_gdf2 = stop_times_gdf[stop_times_col]
-
-    # Merge dfs
-    merge_cols = ["trip_instance_key", 
-                  "schedule_gtfs_dataset_key", 
-                  "shape_array_key"]
-
-    df1 = pd.merge(stop_times_gdf2, trips_to_route_df, on=merge_cols, how="inner")
-    
-    # Fill in missing direction id with 0, per our usual practice.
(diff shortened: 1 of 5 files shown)


gtfs_funnel/schedule_stats_by_route_direction.py
```diff
@@ -12,77 +12,68 @@
...
-        "stop_pair",
-        "stop_pair_name",
-        "stop_primary_direction",
-        "stop_meters",
-    ]
+    STOP_TIMES_FILE = dict_inputs.rt_vs_schedule_tables.stop_times_direction
     
-    # Keep only a subset
-    stop_times_gdf2 = stop_times_gdf[stop_times_col]
+    stop_times_gdf = pd.read_parquet(
```
tiffanychu90, line 21: Add the right indents in this script so that it's not left aligned.

gtfs_funnel/schedule_stats_by_route_direction.py
```diff
@@ -12,77 +12,68 @@
...
                   "schedule_gtfs_dataset_key", 
                   "shape_array_key"]
-
-    df1 = pd.merge(stop_times_gdf2, trips_to_route_df, on=merge_cols, how="inner")
+    
+    stop_times_with_trip = pd.merge(stop_times_gdf, trips_df, on = merge_cols)
     
     # Fill in missing direction id with 0, per our usual practice.
-    df1.direction_id = df1.direction_id.fillna(0)
+    stop_times_with_trip.direction_id = stop_times_with_trip.direction_id.fillna(0)
```
tiffanychu90, line 51: Do you know if this is going to affect `gtfs_digest/merge_data.py`? Does everything merge or do rows drop off? If rows drop off, this `fillna()` should probably be moved into `merge_data`

gtfs_funnel/schedule_stats_by_route_direction.py
```diff
@@ -12,77 +12,68 @@
...
-
+    
     # Drop duplicates so only the top stop_primary_direction is kept.
-    agg3 = agg2.drop_duplicates(
-    subset=[
-        "route_id",
-        "schedule_gtfs_dataset_key",
-        "direction_id",
-    ]).reset_index(drop=True)
+    agg3 = agg2.drop_duplicates(subset= main_cols).reset_index(drop=True)
```
tiffanychu90, line 76: Combine with above and chain! `agg2 = agg1.sort_values(stuff).drop_duplicates(stuff).reset_index()`

gtfs_funnel/schedule_stats_by_route_direction.py
amandaha8: About 745 rows from the `cardinal_direction_for_route_direction()` drop off when it is merged with `schedule_metrics_by_route_direction()`. I experimented with commenting out `fillna()` and 615 rows still drop off from `cardinal_direction_for_route_direction()`.

amandaha8: However, every row from `schedule_metrics_by_route_direction()` merges. There are no `left_only` rows.

Issue #1163, opened 2024-07-01
Research Request - add dwell time for vehicle positions


tiffanychu90:
Complete the below when receiving a research request, and continue to add to this issue as you receive additional details and produce deliverables. Be sure to also add the appropriate project-level label to this issue (eg gtfs-rt, DLA).

## Research Question
**Single sentence** description:  We use vehicle positions, filtering out only trips that are much too short, but do not do additional cleaning on that.
* #592   

**Detailed** description: Dwell time at a location (even if it's not a stop) would be helpful for consolidating vp (lat/lon and timestamp). Store `location_timestamp_local` as the first `vp_idx` observed, capture `dwell_time (seconds)`, and also a `moving_timestamp_local` (the last timestamp at that position).

**Considerations:**
* `dwell_time` would be interesting to look at for stops 
* watch out for: we know that certain stops can be revisited (loopy routes, we might have the same origin and destination). In this case, we care whether the `vp_idx` is increasing by 1 and we want to split non-consecutive ones into separate groups.
Desired grouped table looks like:

| stop  | stop_sequence | vp_idx array for simplicity | arrival | moving |
|-------|---------------|-----------------------------|---------|--------|
| stop1 | 1             | [1, 2, 3]                   | 1       | 3      |
| stop2 | 2             | [4, 5, 6]                   | 4       | 6      |
| stop3 | 3             | [7, 8]                      | 7       | 8      |
| stop1 | 4             | [9, 10]                     | 9       | 10     |

If we do a grouping and do not account for stops being revisited in the same trip (`stop1` would be erroneous and show a dwell time that's extra long):

| stop  | group | vp_idx array for simplicity | arrival | moving |
|-------|-------|-----------------------------|---------|--------|
| stop1 | 1     | [1, 2, 3, 9, 10]            | 1       | 10     |
| stop2 | 2     | [4, 5, 6]                   | 4       | 6      |
| stop3 | 3     | [7, 8]                      | 7       | 8      |

*
* We do not want to overwrite `vp_usable` because we have `rt_vs_schedule` pipeline that needs to look at the minute groupings to see how many `vp_per_minute`. This dwell time table would remove rows and start consolidating rows / removing timestamp observations.
   * However, this dwell time should be used for `segment_speeds` pipeline, and should be used in `vp_condenser` (which condenses vp from long to basically a row (and individual coordinates are strung together in a linestring)

## Metrics
* Dwell time (seconds)
[...]

Issue #1162, opened 2024-07-01
Research Request - Nearest neighbor comparison 


tiffanychu90:
Complete the below when receiving a research request, and continue to add to this issue as you receive additional details and produce deliverables. Be sure to also add the appropriate project-level label to this issue (eg gtfs-rt, DLA).

## Research Question
**Single sentence** description:  Compare the 2 or 3 nearest neighbor vehicle positions selected for speedmaps under Eric's `rt_analysis` and my `gtfs_segments.nearest_snap`. 
* #592 
* this outlines work that supports cleaning up `rt_analysis` modules in #1122 
  
**Detailed** description: Reworked nearest neighbor methodology looks roughly like:
* Find the 1 nearest neighbor vp (vp to select only remove opposite traveling directions, and leaves "Unknown" directions in). Unknown directions mean that bus is at the same lat/lon but has different timestamp.
* The 1 nearest neighbor vp then has up to 2 additional points added (`vp_coords_trio`)
* Idea is that the stop position falls somewhere in the trio of points, either between `point1-point2` or `point2-point3`, and the interpolated position occurs there. If it's out of bounds, then it's either pinned to `point1` or `point3`.
* Within a trip, there is a monotonicity check. All arrival times must be monotonically increasing.
   * If it fails this condition, the interpolated time is set to NaT
   * Within a trip, missing values are now assigned a time based on where the stop is `stop_meters` relative to previous stop and subsequent stop

**Questions / Decision Points**
* How sensitive is speed calculation based on which vp is selected? If we are selecting slightly different vp, like being 1 off, does that matter? In theory, it shouldn't matter much.
* What happens if we select a vp with direction Unknown? How sensitive is it now? Basically, if a bus arrives at a stop and now dwells for 5 min, it would be very sensitive if we select the first point on arrival vs the last point on arrival.
* Can we factor in dwell time and do a smarter speed calculation?
   * Eric's methodology handles this already by asking if the position has moved
   * I do not handle this explicitly (whether the position has moved or not is not known unless positions are looked at sequentially. Need an agnostic way to "know" this)
   * Eric's methodology handles filling in missing stop arrivals with a similar monotonic check, and I do the same. But a bus that's dwelling would have the same `stop_meters`, so....the speed would be probably forced downward? 

| stop  | arrival | dwell | leave | speed_comparison                                           |
|-------|---------|-------|-------|------------------------------------------------------------|
| stop1 | 8:00    | 5 min | 8:05  |                                                            |
[...]

PR #1180, opened 2024-07-18
Add vp dwell time / pilot Big Blue Bus speeds post alignment

tiffanychu90:
## vehicle positions in gtfs_funnel
* Add dwell time (seconds) and create a `moving_timestamp_local` column that captures when that vp starts moving again
* Difference between `location_timestamp_local` and `moving_timestamp_local` is dwell time
* Include this as part of `gtfs_funnel`, and this step should go before `vp_condenser`
   * TODO backfill all dates and make this available
   * TODO update later scripts to use this file / add this file into `gtfs_analytics_data.yml` so the swap is seamless
   * TODO `vp_condenser` should be backfilled too, because this will actually have a lot fewer vp to condense into a linestring
* #1163

## segment speeds
* pilot Big Blue Bus through the rest of the steps to align with Eric's speedmap methodology
* stand up new scripts that only focus on Big Blue Bus to see how pipeline scripts should change
* add `37_bbb_speeds_with_dwell` notebook to quickly look at how Big Blue Bus results. tweak those scripts as necessary, and once it's good, put the content back into the main pipeline scripts.
* #1162 
* change how nearest neighbor outputs are saved. detailed methodology alignment breakdown:

1.  `new_nearest_10`
*  **existing**: nearest neighbor now grabs the one nearest neighbor vp and then grabs 2 points around that (using vp_idx`) and adds all the information about `geometry`, `location_timestamp_local`, and then projects it onto the shape and gets `stop_meters` 
* **alignment**:  grab the nearest 10 vp and save these `vp_idx` out only. grab 10 because there was no guarantee that the 1 point that was grabbed was before the stop. in the absence of knowing, let's grab a large enough list so we can definitely get 1 vp before the stop.
* TODO: backfill all dates once pilot checks out. this overwrites old files.

2.  `new_narrow_to_2`
* **existing**: not an existing step, since we grabbed 1 vp only and populated the trio around it
* **alignment**: add in scheduled shape here, `shapely.project` each of them against the shape. we know where the stop falls on the shape (`stop_meters`), so now we can narrow down to the 2 vp that occur before and after the stop (keep the 2 vp that have the smallest positive and negative deltas from `stop_meters`).
* TODO: backfill all dates once pilot checks out. this creates a new file!

3.  `new_interpolate`
* **existing**: interpolation is done between 2 `location_timestamp_local` columns converted into seconds.
* **alignment**: now that we have dwell time, we will take difference between a `moving_timestamp_local` (previous stop) and a `location_timestamp_local` (current stop). logic should be the time the bus starts moving from previous stop and before it idles at current stop. double check this step.
* TODO: backfill all dates once pilot checks out. this overwrites old files.

4. `quick_bbb_speeds`
* **existing**: no major changes here, simply combine steps in `stop_arrivals_to_speeds` and `average_segment_speeds` into one script for BBB to check results. Save out 2 files to use in exploratory notebooks to figure out how to better filter. Everything in `37_bbb_speeds_with_dwell` is unfiltered so we can look at the full distribution of speeds
* **alignment**: this is our TODO
* TODO: align here, and everything can return into the existing `stop_arrivals_to_speeds` and `average_segment_speeds` scripts

### DIFF
--- a/gtfs_funnel/vp_condenser.py
+++ b/gtfs_funnel/vp_condenser.py
@@ -24,14 +24,15 @@ def condense_vp_to_linestring(
     We will group by trip and save out 
     the vp point geom into a shapely.LineString.
     """
-    USABLE_VP = dict_inputs.speeds_tables.usable_vp
-    EXPORT_FILE = dict_inputs.speeds_tables.vp_condensed_line
+    USABLE_VP = dict_inputs.speeds_tables.usable_vp + "_with_dwell"
+    EXPORT_FILE = dict_inputs.speeds_tables.vp_condensed_line + "_dwell"
     
     vp = delayed(pd.read_parquet)(
         f"{SEGMENT_GCS}{USABLE_VP}_{analysis_date}",
         columns = ["trip_instance_key", "x", "y", 
                    "vp_idx", "vp_primary_direction", 
-                   "location_timestamp_local"
+                   "location_timestamp_local", 
+                   "dwell_time_sec"
                   ],
     )
     
@@ -42,7 +43,9 @@ def condense_vp_to_linestring(
         group_cols = ["trip_instance_key"],
         geom_col = "geometry",
         other_cols = ["vp_idx", "location_timestamp_local", 
-                      "vp_primary_direction"],
+                      "vp_primary_direction",
+                      "dwell_time_sec"
+                     ],
     ).set_geometry("geometry").set_crs(WGS84)
     
     vp_condensed = compute(vp_condensed)[0]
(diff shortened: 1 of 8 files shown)


tiffanychu90:
@edasmalchi: can you look at the files I created for Big Blue Bus that are imported in `37_bbb_speeds_with_dwell.ipynb` and use it in an exploratory notebook? I pulled out some examples, but I'll also look more!

edasmalchi:
yeah!

edasmalchi:
@tiffanychu90 looks like the trip-level speeds are substantially equivalent, but the aggregations vary a lot?

It's not the exact same time period so it's a little hard to compare, but even comparing either peak from the legacy speedmaps to the new combined peak, there's quite a difference. Good news though I think, if the trip speeds are so close it shouldn't be so hard to align the aggregations? Might be worth having separate am/pm in the new pipeline anyway, would allow us to keep presenting more information...

Will continue looking!

https://github.com/cal-itp/data-analyses/blob/eric-bbb-exploratory/rt_segment_speeds/38_bbb_speeds_legacy_compare.ipynb

Issue #1184, opened 2024-07-23
Research Task - patch segment_calcs np.percentile


edasmalchi:
As part of ongoing segment_speeds/speedmaps investigation, noticed that while the individual speeds substantially matched between legacy speedmaps and rt_segment_speeds (as revised in #1180), the aggregated p20/p50/p50 speeds still diverged quite a bit.

Seems like we're inadvertently getting the wrong percentiles based on a difference in how to specify arguements between pd.quantile and np.percentile. [np.percentile](https://numpy.org/doc/stable/reference/generated/numpy.percentile.html), wants 0<q<100, unlike pd.quantile or indeed np.quantile which is 0<q<1. So we seem to be inadvertantly pulling .2, .5, and .8th%ile speeds rather than 20, 50, 80th%ile in `segment_speed_utils.segment_calcs`:

https://github.com/cal-itp/data-analyses/blob/203af07682a07aab8cf239dbe56f90e477786eb5/rt_segment_speeds/segment_speed_utils/segment_calcs.py#L73-L77

tested in [notebook](https://github.com/cal-itp/data-analyses/blob/765bbc402ee8849b4796df50d7d13cbe705f93f5/rt_segment_speeds/38_bbb_speeds_legacy_compare.ipynb) using:

`test = pd.DataFrame({'speed_mph': range(81), 'stop_sequence1': '1'})`

<img width="604" alt="Screenshot 2024-07-23 125353" src="https://github.com/user-attachments/assets/30497f6f-be21-42e5-a1cb-a262d7f489c6">

With this fixed and #1183, it seems like we could have an easy path towards alignment :)

PR #1185, opened 2024-07-23
patch segment_calcs np.percentile

edasmalchi:
see #1184 for context

tested in: https://github.com/cal-itp/data-analyses/blob/eric-bbb-exploratory/rt_segment_speeds/38_bbb_speeds_legacy_compare.ipynb

with this change, speeds and aggregations substantially align for BBB via both legacy speedmaps and segment speeds

Screenshot 2024-07-23 130438
Screenshot 2024-07-23 130454
Screenshot 2024-07-23 130505


tiffanychu90 (approved):
Whoohoo! Lol, never noticed that line at all, great catch!

Issue #1218, opened 2024-08-29
Research Task - add route identifiers to shape-aggregated segment speeds


edasmalchi:
## Research Task

* **Question or Goal**: overarching goal of this issue  

The rt_segment_speeds pipeline currently produces two parallel levels of aggregation: [by shape and by route/direction](https://github.com/cal-itp/data-analyses/blob/main/rt_segment_speeds/scripts/average_segment_speeds.py)

To migrate speedmaps per #1186, I'd like to use the segment speeds data aggregated by shape, but it's also important to have route identifiers (route_short_name, route_id) preserved in that aggregation in order to display route info on the maps and eventual, matching, open data export.

I realize that adding route identifiers to the shape _grouping_ columns could lead to a finer grain of aggregation (generally trips all sharing a route can share multiple shapes, usually at least 2 but there's no guarantee in the GTFS spec that I'm aware of, theoretically trips sharing a shape could share multiple routes).

This seems undesirable. I'd rather propose keeping the grouping columns as-is, and then "aggregate" route identifiers within the groupby (min, max, whatever), keeping the same aggregation grain but with route identifiers that would generally match based on actual practice.

* **Expected Outputs / Findings**: what do you hope to have created when this is complete?

shape-aggregated segment speeds in pipeline now include route identifiers

PR #1229, opened 2024-09-25
speedmap shape segment route identifiers

edasmalchi:
adds route identifiers to speedmap shape segment speeds per #1218


tiffanychu90 (approved):
Yay!

Issue #1230, opened 2024-09-25
Epic - HQTA Methodology Revisions


edasmalchi:
### Summary
* We're having ongoing conversations with HCD and other State agencies and MPOs about HQTA/major transit stop methodology. The goal is to reach a consensus interpretation of the statutes, and align our dataset as closely to that interpretation as possible.
* Additionally, PRC 21155 and 21064.3 have further diverged with the passing of AB2553, we will likely have to adjust our pipeline to treat each somewhat separately.
* I (Eric) continue to prefer our segment/stop-based calculation for hq corridors interpreting PRC21155 "a corridor with fixed route bus service with service intervals no longer than 15 minutes".
* However, PRC21064.3 does specifically include the word "route", so a single-route constraint may be appropriate for intersection-based major transit stops

### Will do
* implement a 20-minute standard _for bus intersection major transit stops_
* revise bus intersection major transit stop buffer to 500ft
* maintain 15-minute standard for hq bus corridors
* move to fixed AM and PM peak periods (6a-9a, 3p-7p), averaging frequencies within each period

### May consider
* only counting frequencies from a single route+direction to meet major transit stop threshold (more likely)
* only counting frequencies from a single route+direction to calculate hq corridors (less likely)

### Notes, misc:
* Most MPOs use 6-9am and 3-7pm as peak periods.
[...]

PR #1252, opened 2024-10-10
2024 HQTA Methodology Revisions

edasmalchi:
Draft PR implementing #1230.

Pending additional review/checks

## Changes implemented

* fixed-duration peaks, with averaging
* 500ft major stop intersection buffer (capture more stops around each intersection, then those get half-mile buffered as before)
* 20min headway standard for underlying major bus stop routes, retain 15min for hq corridors

## Apparent results

* HQ Bus Corridor area decrease from 8.5e+09 -> 5.9e+09 (likely from fixed peak, decrease seems correct)
* Major Bus Stop area increase from 3.6e+09 -> 4.2e+09 (from moving to 20min headway standard, increase seems correct)

## Questions/pending

- [x] Some major stops seemed to "grow" in a ring pattern, even though buffer distance should be the same as before. Will measure. Stops growing in irregular patterns are to be expected (new increased buffer captured distant stop which is now buffered and combined)
- [x] Count of rail major stops seemed to increase?
- [ ] Double-check Amtrak
- [ ] At least one erroneous major stop created where an infrequent route diverges from a frequent route, but the combined stop is the most frequent stop for both route segments. Will explore requiring adjacent segments to meet standard in order to create a major stop... (may spin into second issue, not a blocker in my view)


tiffanychu90:
Re Questions:
 
1. Some major stops seemed to "grow" in a ring pattern, even though buffer distance should be the same as before. Will measure. Stops growing in irregular patterns are to be expected (new increased buffer captured distant stop which is now buffered and combined)

I made an adjustment [here](https://github.com/cal-itp/data-analyses/blob/main/high_quality_transit_areas/assemble_hqta_polygons.py#L101-L108) to more clearly delineate the buffering. I found that when the stop was already buffered from an earlier join, then adding another buffer again convoluted it and I think I left it erroneously buffered for awhile. 
I stamped out the logic to better capture what was desired as: major stop point geom -> buffer by X meters and separately corridor line geom -> buffer by Y meters, and X and Y can now differ.

3. Double-check Amtrak
This [Slack thread](https://cal-itp.slack.com/archives/C02KH3DGZL7/p1708710990047189) kind of gets into a column change that occurs, where Amtrak switched from `regional_feed_type = None` to  `regional_feed_type = Regional Precursor Feed` and gets filtered out.

edasmalchi:
@tiffanychu90 thanks for holding HQTA on the open data run -- since Evan unmarked the Amtrak national feed as a precursor I think I'll merge this and rerun gtfs_funnel + new hqta on Monday since that should pick it up...

Will update the README in the mean time

edasmalchi:
> Count of rail major stops seemed to increase?

Seems intentional per prior refactors -- we used to keep only one row per stop, now we have one per (rail/ferry) stop x route_id.

PR #1259, opened 2024-10-17
Aggregating Metrics on the Operator Grain 

amandaha8:
- Calculating some of the metrics on the operator grain starting from `vp_trip_metrics.`
- Incorporated changes into scripts/MAKEFILE.
- Using this new dataset to create the charts within the GTFS Digest portfolio.

PR #1266, opened 2024-10-24
GTFS Digest & Transit Bunching

amandaha8:
GTFS Digest: addressed Tiffany's comment on my last PR #1259. I incorporated the operator metrics into the existing operator_profiles. I adjusted my GTFS Portfolio code to reflect this. I still have the final dataframe with the suffix AH TESTING, but will fix that once Tiffany checks it out.

Transit Bunching: More timestamp/datetime wrangling. Took the suggestion of only examining 3 routes instead of the entire universe of high frequency routes' data. I am now attempting to create maps and charts to visualize bunching.


tiffanychu90:
Can you rename functions to also follow the same pattern in `merge_data.py`...since the merging script brings now 4 or 5 data sources together, I'd want even the functions to make that obvious, because it's not clear why operator-level metrics and operator stats need 2 functions, they sound equivalent!

* `concatenate_operator_level_metrics` -> `concatenate_schedule_operator_metrics`
* `concatenate_operator_stats` -> `concatenate_rt_vs_schedule_operator_metrics`
* small note: having `merge_cols.remove("organization_name")` 30 lines after, changing what `merge_cols` is so far apart can be confusing. There are 3 usages, 2 of them that are the same (and that's after the list has been re-defined).
   * leading up to this point, only `schedule_gtfs_dataset_key` is present. All the other columns you want to bring in from `mart_transit_database` should be brought in once when you merge in the crosswalk.
   * tidier way, and now there's no need to define and redefine your merge columns: (1) merge schedule with rt_vs_schedule. merge columns are `schedule_gtfs_dataset_key` and date, (2) merge in crosswalk that fills in organization info + NTD info. merge columns are  `schedule_gtfs_dataset_key` and date.

amandaha8:
I'll work on this!

PR #1295, opened 2024-11-18
Updated IIJA README

csuyat-dot:
Updated readme to reflect most recent run of script. Added some more details and context to the existing README


amandaha8:
This looks good to me. You can add that the stakeholders prefer the aggregated version created using `_script_utils.get_clean_data(df, full_or_agg = 'agg')`. I think for step 6, the date format is `MMDDYYYY` without the backslashes.

csuyat-dot:
Addressed comments in latest commit

Issue #1357, opened 2025-01-23
Bug: City of Montebello (Montebello Bus Lines) not showing up in GTFS Digest


evansiroky:
**Where did the bug occur?**
Select from the below, and be sure to affix the appropriate label to this issue (e.g. `dataset`, `jupyterhub`, `metabase`, `analysis.calitp.org`)
- [ ] Data (the warehouse)
- [ ] JupyterHub
- [ ] Metabase
- [x] analysis.calitp.org  
- [ ] Other (add detail)

**Describe the bug**

The City of Montebello (Montebello Bus Lines) not showing up in GTFS Digest - both agency Digest and also District Digest.

**To Reproduce**

[Agency Digest for D7](https://gtfs-digest--cal-itp-data-analyses.netlify.app/district_07-los-angeles):

![Image](https://github.com/user-attachments/assets/cfbbfafb-038b-4dab-80de-3a2e67daa2f5)

[D7 District Digest](https://district-digest--cal-itp-data-analyses.netlify.app/district_07-los-angeles/00__district_report__district_07-los-angeles):

![Image](https://github.com/user-attachments/assets/110833e8-0809-4c2e-b030-21a6255860a1)

**Expected behavior**

- Agency Digest: City of Montebello expected at red line.
- District Digest: City of Montebello expected in red circle.

**Additional context**

Since there was also another reported issue of a missing transit agency in https://github.com/cal-itp/data-analyses/issues/1254, I recommend cross referencing the complete GTFS Digest Agency List with the list of fixed-route agencies on the [Statewide Transit Metrics dashboard](https://dashboards.calitp.org/dashboard/185-statewide-transit-metrics).

amandaha8:
@evansiroky  City of Montebello will show up when I refresh the GTFS Digest with January's data. I cross referenced the GTFS Digest agency list with the 198 fixed-route agencies on the metabase dashboard you linked. While there were 185 matches, there were 14 agencies that were only found in GTFS Digest and 13 agencies that were only found in metabase. Do you believe some of the 13 of the metabase-only agencies should be in the GTFS Digest? 

![Image](https://github.com/user-attachments/assets/34f60f1e-4fde-48ee-a21e-285dd8bbe986)

amandaha8:
[...]

Issue #1312, opened 2024-11-27
Bug: Rail Routes not showing up in map of GTFS Digest


evansiroky:
**Where did the bug occur?**
Select from the below, and be sure to affix the appropriate label to this issue (e.g. `dataset`, `jupyterhub`, `metabase`, `analysis.calitp.org`)
- [ ] Data (the warehouse)
- [ ] JupyterHub
- [ ] Metabase
- [x] analysis.calitp.org  
- [ ] Other (add detail)

**Describe the bug**

Commuter Rail is not showing up on the map of GTFS Digest routes.

**To Reproduce**

1. Go to [Capitol Corridor Page](https://gtfs-digest--cal-itp-data-analyses.netlify.app/district_04-oakland/01__03_report__district_04-oakland__organization_name_capitol-corridor-joint-powers-authority)
2. Scroll to [Map of Routes](https://gtfs-digest--cal-itp-data-analyses.netlify.app/district_04-oakland/01__03_report__district_04-oakland__organization_name_capitol-corridor-joint-powers-authority#map-of-routes)
4. Only the bus route to Auburn shows up.

**Expected behavior**

All routes should show up on the map.

**Additional context**

Some of these long-distance providers might be helpful to show in multiple districts. For example, the Capitol Corridor operates in Caltrans Districts 3 and 4. The specific counties that agencies operate can be obtained in 2 ways. 

1. We try to track this in Airtable via a list of counties the service operates in. 
2. A GIS merge of transit stops to districts could be done.

PR #1370, opened 2025-02-12
Adding back rail routes

amandaha8:
Addresses the issues #1357 and  #1312 
- City of Montebello will appear when I update the GTFS Digest portfolio with January's data.
- Capital Corridor's rail routes are not showing up. I added a section of code in `gtfs_funnel/route_typologies` that adds any rail routes that are filtered out because it doesn't meet the threshold. So far, I have only ran my code in two test files using December and November 2024 data.


tiffanychu90:
I took a look at the 2 notebooks -- great job jumping back through the various stages to debug!
I noticed you said outer joins give too many missing values, but left join was somehow missing rail still? 
I wonder if you can instead use the filtering to always keep rail types (no matter the `pct_typology` value).

```
# earlier in the script, route_dir_cols is defined...but add in `route_type`

rail_types = ['0', '1', '2', '5', '6', '7', '11', '12'] # (i pulled these rail type values from  `tag_rapid_express_rail`)

# If you filter here and keep rail no matter the threshold, wouldn't it always favor rail
route_typology_df = gdf.loc[
   (gdf.pct_typology >= TYPOLOGY_THRESHOLD) | 
   (gdf.route_type.isin([rail_route_types]))
]
        
route_typology_df.to_parquet(
     f"{SCHED_GCS}{EXPORT}_long_{analysis_date}.parquet")
    
# Continue with  on with same functions in the script
primary_secondary_typology(route_typology_df)
categorize_routes_by_name(analysis_date)
      
```

The reasoning in simplifying the structure here is: 
* `route_typology_df` is the filtered df for values we want
* Pass through 2 functions: `primary_secondary_typology` and `categorize_routes_by_name`
* Now rail routes are grabbed and passed through `primary_secondary_typology` and `categorize_routes_by_name` again 
   * Can we move away from nesting those 2 functions? If you always want rail, then grabbing it regardless of `pct_typology` makes sure you always do so, rather than patching it in after the fact.
* concatenate part1 and part2 dfs for your final

Issue #1372, opened 2025-02-13
Research Request - Time-of-day grains for speeds + refactor segment speed averaging


tiffanychu90:
Complete the below when receiving a research request, and continue to add to this issue as you receive additional details and produce deliverables. Be sure to also add the appropriate project-level label to this issue (eg gtfs-rt, DLA).

## Research Question
**Single sentence** description:  We want to aggregate across time for segment speeds. Currently, we have a segment-trip grain table that is created, and that's used directly in `average_segment_speeds`.  Add one more output that's segment-time_of_day grain that is more flexible across various aggregations across time.
* Further develops #1202 

**Detailed** description:

- [x] Create time-of-day aggregation, backfill for all dates
   - [x] stop segments #1373, #1374 
   - [x] rt stop times #1410 
   - [x] speedmap segments #1410 
- [x] Can we consolidate the segments in a way that backfilling is easier? At this point, 2023-2025 is quite a lot
   - [x] Consolidate under `rt_stop_times`. Move the aggregations desired from `stop_segments` into `rt_stop_times` - not exactly equal, but with `route-direction-stop_pair`, it should be close enough. Done in #1406.
   - [x] Backfill the new output for past dates using `speeds_by_time_of_day` scripts. Done in #1410 
- [x] Use better keys in `gtfs_analytics_data.yml` to make clear the grain we're creating.
   - [x] `rt_stop_times`: new keys created in #1406  
   - [x] `stop_segments`: deprecate in #1411
   - [x] `speedmap_segments`: new keys created in #1408
- [x] Implement a weighted average approach? Treat each day as independent entity that gets brought together across time.  Done in #1408
   - Reference: https://github.com/jsvine/weightedcalcs
- [x] Refactor `average_*_speeds.py` (3 sets of scripts) to split out single day vs multiple day. Each script is being used to handle 2 types of aggregation...it should just do a single day aggregation and we can use a separate script to handle multiple day aggregation. Done in #1408
   - `speeds_by_time_of_day.py`: handle single day time-of-day aggregations
   - `average_summary_speeds.py`: handle single day summaries for route-direction speeds
   - `average_segment_speeds.py`: handle multiple day aggregations. Uses single day time-of-day aggregation and aggregates up to weekday/weekend-year-time-of-day using weighted averages. 
[...]

PR #1402, opened 2025-02-27
Refactor averaging segment speeds + Feb 2025 segment speeds pipeline

tiffanychu90:
# segment speeds
* Start with checking whether the new `fct_vehicle_locations_grouped` gives us similar results as before
  * notebook `49_foothill_new_mart.ipynb` shows there are differences even in the stage2 dataset
  * revert #1392 - reinstate the vp processing scripts until we can figure out why the upstream swap is not equivalent
  * save out  Jan 2025 `_new` for `rt_stop_times`: `stage2, stage3, stage4` to see where the differences are
* Refactor how averaging of segment speeds occurs
   * Address the challenges: segments are granular, and throwing a week at it can crash the kernel with `dask.delayed`, but we really just want to aggregate before computing. Use `dd.from_map`, which handles it better, add grouping columns, aggregate, then compute.
   * Another challenge is how to bring in segment geometry again when we have multiple days. Instead, we'll use `dd.from_map` and dedupe, then use the `schedule_gtfs_dataset_key-route-direction-stop_pair` to merge it back on (there might still be dupes, but this sufficiently pares it down to attach) 
* Consolidate the segment types under `rt_stop_times` as the main one
   * If there are changes in upstream tables, rerunning all the dates from 2023-2025 can be quite a lot
   * Add the outputs that were under `stop_segments` under `rt_stop_times` and get similar `route-direction-stop_pair` aggregations that are going to be close enough. `stop_pair` solves a big part of how trips can use different stop_sequences but refer to the same stop, so if we can use it, we don't need to try and solve the variety of shapes that serve a route-direction.
   * segment types use `segment_timeofday` as the basis for building all other aggregated datasets 
   * New keys in `gtfs_analytics_data.yml` reflect the grains for aggregation
 *  Update `segment_calcs` to handle weighted averages - checked that these are equivalent to reading in all segment-trip speeds   
* #1372

# open data
* Run Jan 2025 and Feb 2025 again for `gtfs_funnel` + `rt_segment speeds`. Make sure these are the same methodology as before.
* Run `rt_vs_schedule` pipeline

### deprecation
Remove scripts that have now been refactored and consolidated 
* `average_speedmap_segments.py` --> handled in `speeds_by_time_of_day.py`, which is part of the normal pipeline
* `quarter_year_averages.py` --> `average_segment_speeds.py` now handles multiple day averages
* remove `time_of_day_averages.py` --> moved into `speeds_by_time_of_day` and handles all the segment types
* comment out `stop_segment` related scripts under `rt_segment_speeds/scripts/Makefile`, and once the new outputs are created, these will get removed 
   * Add the refactored `average_stop_segment_speeds` script before we decided to consolidate version


tiffanychu90:
Close - too many issues tackled...break out into smaller PRs

PR #1413, opened 2025-03-07
GTFS Digest

amandaha8:
- Buffered CA shape so the bay is included as part of the state. Now, several ferry operators that were previously erased are back.
- Revisited the old issue of why Metrolink/San Bernardino are missing and still categorized under "realtime only": their trips.txt file doesn't include` shape_id`. Evan's team will follow up with them.
- Revisited issue of one `schedule_gtfs_dataset_key` to multiple `organization_name` values. We don't want to have duplicated `schedule_gtfs_dataset_key` information shown under different operators on our portfolio. I added a new script in `_shared_utils` that creates a yml with because we want a source of truth file that will show these `schedule_gtfs_dataset_key`  examples with a 1:m relationship. 
- Updated the script that creates the `yaml` behind the GTFS Digest so only a `schedule_gtfs_dataset_key` will only be featured once under one `organization_name` .


tiffanychu90:
I think your next PR can handle this yaml. As is, the first `schedule_gtfs_dataset_key` is linked to 3 organizations, yes? 

I would like to see where there are multiple organizations, which was the organization name selected.

```
# Current set up
  - caption: 01 - Eureka (a253a8d7acd57657bb98050f37dd6b0f)
    params:
      District/Key: 01 - Eureka (a253a8d7acd57657bb98050f37dd6b0f)
    sections:
    - organization_name: City of Arcata
    - organization_name: City of Eureka
    - organization_name: Humboldt Transit Authority
```

But instead:
```
# if you want to swap the order of key and value, that's fine.
# I put organization_name as the key because you are likely to start from the digest and check it against
# the schedule_gtfs_dataset_name we use for the feed
# I removed the nested structure - is it important for it to be organized by district or is alphabetical enough?

organization_name: schedule_gtfs_dataset_name (sorted alphabetically on both key and value)

Foothill Transit: Duarte Schedule
Foothill Transit: Foothill Schedule
LA Metro: LA Metro Bus Schedule 
LA Metro: LA Metro Rail Schedule


```

tiffanychu90:
Also - make sure to use the `precommit` commands, which are active for `shared_utils`....listed in [the Makefile](https://github.com/cal-itp/data-analyses/blob/main/Makefile#L76-L79) and [written out in the docs](https://docs.calitp.org/data-infra/contribute/submitting_changes.html#making-changes-and-merging-prs). 

In the follow-up PR where you adjust the yaml, make sure it passes precommit hooks, similar to when you update the docs site in `data-infra`

PR #1418, opened 2025-03-12
jan 2025 monthly ntd ridership report publish 

csuyat-dot:
epic:
- #1027 

Ran monthly ntd ridership report script for jan 2025. deployed notebooks to netlify. ensured data was exported to public/private GCS buckets. did not encounter any errors.


csuyat-dot:
was initially failing pre-commit checks, ran `per-commit run --all-files` and it identified a script that needed some manual changes. pushed a new commit with those changes and now everything passes.

Issue #1423, opened 2025-03-18
Research Request - Conveyal Network Update


hhmckay:
## Research Question
Update the underlying GTFS and OSM networks in Conveyal.
  
Caltrans staff use the Conveyal tool to calculate accessibility metrics. The tool relies on underlying GTFS and OSM datasets to calculate trip times for walk, bike, pedestrian, and transit modes. These datasets need to be updated periodically to represent current transit/street networks. The last update was performed over a year ago so updated data is needed.

## How will this research be used?
This updated data will be used by analysts to score projects on three accessibility-derived metrics for the ITIP program, as well as other upcoming accessibility analysis efforts.

## Stakeholders & End-Users
Henry McKay, Director's Office of Sustainability

## Data sources
* **Cal-ITP data sources**:
* https://github.com/cal-itp/data-analyses/tree/main/conveyal_update

## Deliverables
Updated Conveyal networks for all four state regions.

## Timeline of deliverables
To be completed by 3/28 if possible.

KatrinaMKaiser:
Hey @hhmckay what's the significance of the 3/28 deadline? Do you have any flexibility on it?

edasmalchi:
I said I could get it done by then, since I’ll be out the week after.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

hhmckay:
Pretty flexible. Just want to get it done before we start running ITIP projects. I spoke with Eric and he said next week is doable.

mrtopsyt:
All network bundles should be added to Conveyal now! Thanks for both of your help!

KatrinaMKaiser:
@mrtopsyt is this issue closeable?

mrtopsyt:
I linked the PR and am closing this now. Apologies for the delay, I didn't have my Github notifications set up properly at this time so I missed your comment, they seem to be working now though so it shouldn't be an issue going forwards.

PR #1440, opened 2025-04-07
Conveyal Update + Automation

mrtopsyt:
- Ran Conveyal update for March 2025
- Added check to avoid downloading non-customer-facing feeds that are not regional subfeeds
- Added feature to search for past feeds that were removed and replaced with a feed that was only valid after the target date, making them invalid on the specified service date


edasmalchi:
@mrtopsyt this closes #1423, right?

mrtopsyt:
I think so! I closed the issue, sorry about that @edasmalchi

edasmalchi:
all good, thanks Anna!

PR #1451, opened 2025-04-15
NTD Transit Facilities Map

mrtopsyt:
- Added a notebook and associated code in `ntd/ntd_transit_facilities` showing data from the 2023 NTD Transit Facilities Inventory
  - Geocoded addresses where lat/lon was not provided
  - Created folium map with color coding to display results
  - Support caching geocoded results in GCS as Shapefiles and Geojson
- Data is available on [GCS](https://console.cloud.google.com/storage/browser/calitp-analytics-data/data-analyses/ntd/ntd_transit_facilities?invt=AbuxLQ&project=cal-itp-data-infra&pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22)))


csuyat-dot (approved):
Looks good to me!

mrtopsyt:
Thanks! TDQ has been asking about using this [geocoded version of the Facilities Inventory](https://data.transportation.gov/stories/s/gd62-jzra), since it means we're not responsible for geocoding errors, but it has a lot of data cleaning issues that aren't present in the non-geocoded version. I'm merging this now, later down the line we might need to look into that if this product becomes used for the GTFS parent stops issue and needs better QC.

PR #1472, opened 2025-05-08
does not raise error if no feeds from alternate dates are found

mrtopsyt:
- Fixed error in the Conveyal network bundle download where `evaluate_feeds.py` would raise an error if no feeds from alternate dates were found to replace feeds unavailable on the specified service date.


edasmalchi (approved):
lgtm!

PR #1478, opened 2025-05-12
compare ntd_ids from mart_ntd against dim_orgs

csuyat-dot:
When comparing the new RTPA column in `dim_organizations` to the rtpa column in the `ntd_id_rtpa_crosswalk_all_reporter_types`, derived from `mart_ntd_funding_and_expenses.fct_service_data_and_operating_expenses_time_series_by_mode_upt` (mar_ntd), we found there some unique agencies in mart_ntd that do not appear in dim_orgs.

38 agencies from mart_ntd were check against dim_orgs to find any equivalent agencies. Goal was to add any missing ntd_id or RTPA data to dim_orgs or add entirely new agencies. This is to ensure dim_orgs has the most up to date ntd_id and RTPA values to use for analyses, instead of using the older crosswalk. 

6 scenarios were identified to explain the different combinations of agency name and ntd_id that were observed and explained in the notebook.


csuyat-dot:
still need to determine what to do with the remaining agencies that did not match to a dim_org agency. Should they be added to airtable? if so, then need to also assign the RTPA for the agency.

csuyat-dot:
Teams message i sent in office hours
>Hi all, finally have an update regarding these the distinct agencies i found in mart_ntd_funding_and_expenses.fct_service_data_and_operating_expenses_time_series_by_mode_upt (mart_ntd) table that do not exist in dim_organizations. I detailed my findings here https://nbviewer.jupyter.org/github/cal-itp/data-analyses/blob/4c34b719f1f6fe7681b312c3fc366cdcde2f32a0/ntd/qc_xwalk_dim_orgs.ipynb
> 
>tl;dr. some agencies had an equivalent name matches between both tables, so their ntd_ids were copied over to ntd_id_2022 in airtable. Most agencies however, did not have a name match so we need guidance on how to proceed; should these orgs be added to airtable or not, can orgs be added as alias?

|                      Scenario                     |                                                        description                                                       | Row Count |                                                                                                       Example(s)                                                                                                       |                                                            Result / Question                                                           |
|:-------------------------------------------------:|:------------------------------------------------------------------------------------------------------------------------:|:---------:|:----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------:|:--------------------------------------------------------------------------------------------------------------------------------------:|
| 2.   name: match \| dim_org.ntd_ID: different     | Where the mart_ntd and dim_org   agency name are the same, but the dim_org ntd_id is different                           | 12        | - "Los   Angeles County Metropolitan Transportation Authority (LACMTA) 90021" vs "Los Angeles County   Metropolitan Transportation Authority 90154"     - "Paratransit, Inc. CTSA 90224" vs "Paratransit Inc.   90223" | No changes made due to existing   NTD_ID. Can mart_ntd agency names be added as aliases in   Airtable?                                 |
| 4. name:   none \| dim_org.ntd_ID: none           | Where the mart_ntd name does not   have a similar, or any equivilent name, in dim_org. thus, no dim_org ntd_id   either. | 6         | - "ATC   / Vancom 90170" vs "no   name, no id"- "Laidlaw   Transit Services 90178" vs "no name, no id"     - "DAVE Transportation Services, Inc.   90158" vs "no   name, no id"                                        | No changes. Agencies likely out   of business (last reported to NTD: 2001–2015). Should   these be included in Airtable? Probably not. |
| 5. name:   different \| dim_org.ntd_ID: different | When the mart_ntd and dim_org   name have different, but slightly similar, names. with different ntd_id.                 | 2         | - "LACMTA   - Small Operators (LACMTA) 90166" vs "90154 Los Angeles County Metropolitan Transportation   Authority"     - "MTS Contract Services (MCS) 90185" vs "San Diego Metropolitan   Transit System 90026"       | No changes. Can   they be added as aliases?                                                                                            |
| 6. name:   different \| dim_org.ntd_ID: none      | Where mart_ntd and dim_org names   are different, but slightly similar. but no ntd_id in dim_orgs                        | 7         | - "Chico   Area Transit System City of Chico (CATS) 90128"   vs "City of Chico n/a"     - "Paso Robles Transit Services (PE)   90195" vs "City   of Paso Robles n/a"                                                   | Some names matched clearly →   NTD_ID copied over. Others names were unclear → No changes made.                                        |

csuyat-dot:
via Evan:
>scenario 2. Can we do some outreach to these agencies to figure out why they have multiple NTD IDs?
>scenario 4. If we don't need them, then yeah we probably don't need to add them.
>scenario 5. Same question as 2
>scenario 6. When there is a difference of names, we may need to ask what is your unit of analysis that you need? Sometimes agencies have legal names, sometimes their services have different names.

Issue #924, opened 2023-10-05
Research Request - Route identification over time


tiffanychu90:
Complete the below when receiving a research request, and continue to add to this issue as you receive additional details and produce deliverables. Be sure to also add the appropriate project-level label to this issue (eg gtfs-rt, DLA).

## Research Question
**Single sentence** description:  We want to identify the same route over time, despite small changes in `route_id`, `route_long_name`, `route_short_name`, and `route_description`. 
* First noticed this in #806, and now this is something we want to address before doing more time-series work in schedule or RT
  
**Detailed** description: Using GTFS schedule `routes`, we have `route_id`, `route_long_name`, `route_short_name`, and `route_description`. Once we started aggregating for the month, and even across months, `route_id` in itself is not that stable (if suffixes change). 

## How will this research be used?
* We want some logic implemented to identify similar enough characteristics across the 4 route columns to know that they are indeed the same route.
* Update this function in [portfolio_utils](https://github.com/cal-itp/data-analyses/blob/main/_shared_utils/shared_utils/portfolio_utils.py#L59-L78) and [rt_utils](https://github.com/cal-itp/data-analyses/blob/main/_shared_utils/shared_utils/rt_utils.py#L731-L739) later

## Data sources
* **Cal-ITP data sources**: 
* Either cached `trips` data that is already downloaded or `mart_gtfs.dim_routes` and just grab it all at once
* Start with `{COMPILED_CACHED_VIEWS}trips_{analysis_date}.parquet` and concatenate a couple days from `shared_utils.rt_dates` together

## Deliverables:
Notebook, and/or script or function

PR #1526, opened 2025-06-17
add route identification notebook

tiffanychu90:
* Review how routes are getting standardized in time-series.
* VCTC could benefit from being removed from the list in `time_series_utils`: [operators_only_route_long_name](https://github.com/cal-itp/data-analyses/blob/9505ad4445de05119b78d4ad35ac89c23ed8a7dc/rt_segment_speeds/segment_speed_utils/time_series_utils.py#L84-L105) based on their more recent changes. Potentially others?
* #924

### DIFF
--- a/gtfs_funnel/08_route_identification.ipynb
+++ b/gtfs_funnel/08_route_identification.ipynb
@@ -1,0 +1,317 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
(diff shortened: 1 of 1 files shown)


tiffanychu90:
@amandaha8: can you run the notebook and come up with a list of which ones you'd change? I set up a quick exploratory to make the decision around VCTC, but haven't taken a closer look related to the other operators, though I quickly spotted 1 other one that could also do removing the extra parsing.

Issue #1528, opened 2025-06-17
Research Task - GTFS Funnel + Open Data for June 2025


mrtopsyt:
## Research Task

Run the GTFS Funnel and Open Data Publishing Workflows for `2025-06-11`

* **Relevant PR** (if applicable):  

* **Question or Goal**: Update Open Data and data from the GTFS Funnel

* **Data Required**:
  * Currently have: GTFS Warehouse Data
  * Need:  NA

* **Research Required**:  NA

* **Metrics**: what specific calculations are you looking to find?  

* **Expected Outputs / Findings**: what do you hope to have created when this is complete?

- [ ] Run the GTFS Funnel
  - [x] Update dates
  - [ ] Run Funnel
- [ ] Generate Open Data Shapefiles
- [ ] QC Open Data Shapefiles
  - [ ] Metadata
  - [ ] Consistent row counts
  - [ ] Consistent names
- [ ] Upload Open Data Shapefiles and notify DRISI
- [ ] Update Funnel Readme
- [ ] Update Open Data Readme with any changes

PR #1538, opened 2025-06-25
Split vp preprocessing out

tiffanychu90:
* Split the `vp_direction.py` out again from `vp_keep_usable.py` for #1528
* Adjust `Makefile`, add logs
* @mrtopsyt: you can cherry pick these and work it into your larger PR
   * I did notice `cleanup.py` wasn't run for a couple months in 2025, there are a couple of staging files left from March and April 2025
   * I also noticed May 2025 was missing stuff from the `vp_preprocessing` stages, so you should probably go and run those


tiffanychu90:
Close as #1545 is merged in

Issue #1525, opened 2025-06-17
Research Task - Monthly NTD, refactor code


csuyat-dot:
## Research Task

Substantial updates to the Monthly NTD ridership report necessitate a review and possible refactor of the report's code. Since the initial iterations of the report, improvements have been made to; ingesting data from the warehouse instead of the FTA excel files, ingesting rtpa data from the warehouse instead of a curated crosswalk.

Now is a good time to deprecate or entirely remove old functions, and general code clean up. 
ex. Are `update_vars` or `ntd_dates` modules still needed? Can we consolidate functions from the other products in the same directory

* **Relevant PR** (if applicable):  

* **Question or Goal**: overarching goal of this issue  

* **Data Required**:
  * Currently have:
  * Need:  

* **Research Required**:  

* **Metrics**: what specific calculations are you looking to find?  

* **Expected Outputs / Findings**: cleaner and more organized code.

csuyat-dot:
consolidate and move the common functions  to the top utils files so it can be used with the other products. 
- `deploy_portfolio_yaml`
- `remove_local_outputs`
- `save_rtpa_outputs`
- `sum_by_group`
- `add_change_columns`
- `get_percent change`

csuyat-dot:
have a single `Makeffile` at /ntd with different sections for each report to consolidate all the other Makefiles?

PR #1544, opened 2025-06-30
Refactor Monthly and Annual NTD Ridership sites, part 1

csuyat-dot:
issue
- #1525 

Consolidated common functions between both sites to `_01_ntd_ridership_utils`. Refactored functions to work with both reports (added new arguments, options to account for report specifics columns. Tested updated functions in explore notebook. Tested update functions in the individual report folders. Confirmed each report script still produces data, produces output files, and that notebooks can still run.


KatrinaMKaiser (commented):
If you remove the deprecated code rather than commenting it out, it will be easier to read.

ntd/monthly_ridership_report/monthly_ridership_report.ipynb
KatrinaMKaiser, line 1: This might be out of scope for the current refactor, but a future step might be defining more functions in an outside script

Issue #1560, opened 2025-07-10
Research Request - IIJA Building CA Map


amandaha8:
## Research Question
**Single sentence** description:  
The Building CA team requested updates to the script that prepares the data for the Building CA map. 

**Detailed description:** 
Take the raw data and run it through our script. Solve any bugs that come up. 

## How will this research be used?
Refresh the Building CA map with new data. 

## Stakeholders & End-Users
Building CA team.

## Deliverables
- Updated and cleaned CSV of data.

PR #1567, opened 2025-07-15
IIJA Data

amandaha8:
- Cindy Tsai requested we run the raw data through the script to generate clean data for the Building CA Map.
- Removed some parts of siuba from the code.
- The file we use to incorporate county information changed, so the code was updated with the new column names. 
- Light code cleaning.


csuyat-dot:
issue:
- #1560

Issue #1547, opened 2025-07-03
Chore - Use the new GCS auth helpers where possible


raebot:
Replace calls to `read_parquet`, `read_file`, and `geo_data_frame.to_parquet` with GCS helpers introduced in [calitp-data-analysis v2025.6.24](https://github.com/cal-itp/data-infra/pull/4004) so it's easier to access these resources by virtue of not having to fetch a token each time. 

**Detailed** description:

## Stakeholders & End-Users
Analysts' lives will be made easier by replacing inline GCS auth token or filesystem calls with these helper functions because there will be a clear and consistent way of accessing these geospatial resources.

KatrinaMKaiser:
Yes I think so, it's a fundamental and wide ranging improvement

PR #1574, opened 2025-07-19
Demonstrate use of GCS auth helpers

raebot:
This PR demonstrates use of the new GCS auth helpers `read_parquet`, `read_file`, and `geo_data_frame_to_parquet` in a couple files. It removes the need to import `gcsfs` or `google.auth` packages and manage auth tokens. The idea here is to provide a few examples and then let analysts begin using these helper functions themselves as they find them useful. I also optimized imports in the files I edited while I was in there.

This PR relates to the helper introduced with [this `calitp-data-analysis` PR](https://github.com/cal-itp/data-infra/pull/4004/files#diff-3ffdac9594e95794cbe0d47e9a30617e20e02eb9294b7410c16991b7f503be10).

Resolves https://github.com/cal-itp/data-analyses/issues/1547

### DIFF
--- a/gtfs_digest/_transit_routes_on_shn.py
+++ b/gtfs_digest/_transit_routes_on_shn.py
@@ -1,30 +1,24 @@
+from functools import cache
+
 import geopandas as gpd
 import numpy as np
 import pandas as pd
+from calitp_data_analysis.gcs_geopandas import GCSGeoPandas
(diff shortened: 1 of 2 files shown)


amandaha8:
@raebot, I tried to run the script you updated in my terminal, but I received the result "killed." 
image

Issue #1514, opened 2025-06-13
Research Task - Incorporate feedback to Transit Performance Metrics Portfolio


csuyat-dot:
## Research Task

Incorporate the feedback received from Juan Matute of UCLA ITS. Specially to address:
>If you're taking requests, I'd like to see the Table 8.1 performance metrics on a statewide basis, along with a list for each performance metric of which individual transit agency-mode of service combinations are in the bottom 5% (approximately two standard deviations from the mean) for each. This would be illustrative for discussion purposes.
>
>~~In a more advanced version, operators would be clustered into groups of 10 or more based on mode of service, vehicles available, population density of service territory, job density of service territory, and, perhaps, service area overlap with other transit operators (a GTFS spatial analysis exercise). An agency scoring in the bottom 1 or 2 of the cluster would get some remedial help in their triennial audit. Or face consolidation (FWIW, I like the [BC Transit](https://urldefense.com/v3/__https:/en.wikipedia.org/wiki/BC_Transit*History__;Iw!!LWi6xHDyrA!4EoPfB66YZt7S5OKCHStMrbV7mwmBsEWnq3dOVlEnrfXC5wi8H_AlpHColsZuJjBb4usrvlh-4lSw6dkdqjhBHXj$) model for consolidation starting in 1979). Several large transit operators, especially those operating rail, wouldn't be candidates for consolidation and wouldn't fit this clustering method and would instead rely on a triennial audit, where I would expect trends over time for GTFS-RT quality, customer experience metrics (Transit App surveys or mystery shops) and several of these metrics to be considered holistically.~~
>
>And perhaps agency costs would be adjusted for [regional](https://urldefense.com/v3/__https:/www.dir.ca.gov/oprl/capriceindex.htm__;!!LWi6xHDyrA!4EoPfB66YZt7S5OKCHStMrbV7mwmBsEWnq3dOVlEnrfXC5wi8H_AlpHColsZuJjBb4usrvlh-4lSw6dkdvr8eD6_$)[ consumer price index](https://urldefense.com/v3/__https:/www.dir.ca.gov/oprl/capriceindex.htm__;!!LWi6xHDyrA!4EoPfB66YZt7S5OKCHStMrbV7mwmBsEWnq3dOVlEnrfXC5wi8H_AlpHColsZuJjBb4usrvlh-4lSw6dkdvr8eD6_$) maintained by California Department of Industrial Relations. Either that or they'd just be clustered with regional peers.

[...]

PR #1581, opened 2025-07-23
Statewide Transit Performance Metrics Analysis

csuyat-dot:
issue:
- #1514 

Read in data used for transit performance metrics by RTPA, but adjusted to see all agencies in the state. Separated performance metrics into 2 groups: cost-effectiveness and service effectiveness. Per the issue, also separated out the bottom 5% of agencies for each performance metric.


ntd/new_transit_metrics/explore_statewide_transit_metrics.ipynb
KatrinaMKaiser, line 1: Regarding getting rid of rows with 0 values, how many agencies/services are affected by this? print them out. I think you should calculate the performance metrics and then get rid of the agencies with inf or NA values (and display those), rather than getting rid of the rows with any 0 in the raw numbers.

KatrinaMKaiser (changes requested):

ntd/new_transit_metrics/explore_statewide_transit_metrics.ipynb
csuyat-dot: Latest commit should address these points! [link to latest nbviewer](https://nbviewer.org/github/cal-itp/data-analyses/blob/juan_metrics/ntd/new_transit_metrics/explore_statewide_transit_metrics.ipynb)

1. In cell 1, performance metrics are calculated first. inf/NaN values were identified in cell 3.
2. Cells 4-6 compares filtering the dataframe by either: removing zero values or removing inf/NaN values. Both method produce equivalent results
3. Cell 7 is a cleaner view of the rows to be excluded. 11 rows total: 7 rows with at least 1 zero value, 4 rows with ALL zeros. Each row resulted in inf or NaN performance metrics.

PR #1583, opened 2025-07-24
Update portfolio index on change

jparr:
Update portfolio/index.html when merging changes

Related: #3976


jparr:
Example run here: https://github.com/cal-itp/data-analyses/actions/runs/16479617900/job/46605915520

PR #1584, opened 2025-07-24
Pilot migrating notebook off of siuba

raebot:
This PR migrates the `ntd/new_transit_metrics/explore_new_transit_metrics` notebook off of siuba and to instead use SQLAlchemy and pandas directly. I already did one pass at verifying that outputs were equivalent for each migrated query, but will need @csuyat-dot to confirm as the owner of the notebook. This migration is done as an example that can be followed for migrating other code that uses siuba.


csuyat-dot (approved):
Compared latest outputs via sqlalchemy/pandas to initial outputs via siuba. Confirming that outputs are equivalent!

PR #1618, opened 2025-08-14
1600 backfill rta tests

mrtopsyt:
- Created tests as follows for `make_retrospective_feed_single_date`
  - Test whether a feed has all tables from `filtered_input_feed`
  - Test whether all tables from `filtered_input_feed` except stops.txt and stop_times.txt are unchanged in the output
  - Test that, if there are trips in `stop_times_table` that are not in the schedule, a key error is raised
  - Test that trips that are only present in `filtered_input_feed` are dropped from the output trips table
  - Test that trips that are only present in `filtered_input_feed` are dropped from the output stop times table
  - Test that `arrival_time` and `departure_time` in the output `stop_times` match `stop_times_table` where `trip_id` and `stop_sequence` are equal
  - Test that all attributes in the output are not the same object as the input in `filtered_input_feed` (to ensure that there aren't mutability issues)
  -  Test that if `validate=True` and `filtered_input_feed` and `stop_times_table` do not have matching stop ids at the same value of trip_id and stop_sequence, an `AssertionError` is raised
  - Test that if `validate=True` and `filtered_input_feed` and `stop_times_table` do not have matching schedule arrival times at the same value of trip_id and stop_sequence, an `AssertionError` is raised
  - Test that, if the data in `filtered_input_feed` and `stop_times_table` represent identical trips, the `stop_times` table in the output feed is identical to the schedule version except that columns are as specified in `stop_times_desired_columns`
- Make fixes in `make_retrospective_feed_single_date` where these tests were failing


realizable_transit_accessibility/tests/test_retrospective_feed_generation.py
```diff
@@ -0,0 +1,335 @@
...
+
+
+def test_trips_present_in_rt_only_leads_to_exception(
+    schedule_feed_minimal: GTFS, two_trip_rt_table_schedule_rt_different: pd.DataFrame
+):
+    """Test that, if there are trips in `stop_times_table` that are not in the schedule, a key error is raised"""
+    # Check setup conditions
+    assert (
+        schedule_feed_minimal.trips.trip_id == 1
+    ).all(), "Missing or extraneous trip ID in schedule feed (inficates bad test setup)"
```
raebot, line 140: I understand this is intended to test that the `trip_id` for each of the trips equals 1. Right? The syntax is throwing me off a bit. Can you link me to a documentation page that covers this usage? I'm familiar with the Python `all([1,2,3,4])` builtin function, but not this chaining style. Is it Pytest specific?

realizable_transit_accessibility/tests/test_retrospective_feed_generation.py
```diff
@@ -0,0 +1,335 @@
...
+        filtered_input_feed=two_trip_schedule,
+        stop_times_table=two_trip_rt_table_schedule_rt_different,
+        **DEFAULT_TEST_FEED_GENERATION_KWARGS,
+    )
+
+    # Check output
+    # Check that both expected trips are present and there are no new trip ids
+    assert 1 in output_feed.stop_times.trip_id
+    assert 2 in output_feed.stop_times.trip_id
+    assert output_feed.stop_times.trip_id.isin([1, 2]).all()
```
raebot, line 226: You only really need this third assertion right? It makes the first two redundant.

realizable_transit_accessibility/tests/constants.py
```diff
@@ -0,0 +1,58 @@
...
+TEST_DEFAULT_COLUMN_MAP = {
+    SCHEDULE_ARRIVAL_SEC: SCHEDULE_ARRIVAL_SEC_NAME,
+    RT_ARRIVAL_SEC: RT_ARRIVAL_SEC_NAME,
+    STOP_SEQUENCE: STOP_SEQUENCE_NAME,
+    TRIP_INSTANCE_KEY: TRIP_INSTANCE_KEY_NAME,
+    TRIP_ID: TRIP_ID_NAME,
+    STOP_ID: STOP_ID_NAME,
+    SCHEDULE_GTFS_DATASET_KEY: SCHEDULE_GTFS_DATASET_KEY_NAME,
+    RT_ARRIVAL_SEC: RT_ARRIVAL_SEC_NAME,
+}
```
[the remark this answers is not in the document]
mrtopsyt: in hindsight, I'm not sure either. When I initially wrote the feed generation code, I thought it would be good to specify the columns in the schedule rt table through this `stop_times_table_columns` parameter. This was because I was (and am still) unsure how how stable the column names in gtfs funnel products are, and because in theory it would be useful if someone else wants to run this with their own RT stop time table they got from another source (one of the reasons I have to write this is because other open-source attempts to do retrospective accessibility were hardcoded to use APIs to get RT times that no longer exist). 

I'm not sure if this "identity" map was the best way to do it though. For now, I've changed it so that `make_retrospective_feed_single_date` takes `DEFAULT_COLUMN_MAP` as a default argument, which I think makes it a lot simpler, but I'm not sure if there's a better way to handle this entirely.

realizable_transit_accessibility/tests/test_retrospective_feed_generation.py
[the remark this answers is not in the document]
mrtopsyt: Thanks for catching those🙃

realizable_transit_accessibility/tests/test_retrospective_feed_generation.py
raebot: Oh, I see. Makes sense now.

PR #1623, opened 2025-08-14
Removing Siuba

amandaha8:
- Removed files from `gtfs-digest`, `gtfs-schedule`, `rt-vs-schedule` that have siuba references and are no longer used.
- Deleted out siuba from the import statements that are unused.
- Refactored the `dla/iija` work that relies on `siuba` and made sure the script ran completely.


KatrinaMKaiser:
Amanda would you say the rt_scheduled_v_ran stuff is still being maintained, or can it be archived?

Issue #1591, opened 2025-07-31
Research Task - Create `raw_feed_download_utils` package to allow access to `conveyal_update` functions


mrtopsyt:
## Research Task - Create `raw_feed_download_utils` package to allow access to `conveyal_update` functions

Currently, `conveyal_update` contains useful functions to download GTFS feeds from GCS and obtain necessary metadata. Refactor functions from `evaluate_feeds.py`, `match_feeds_regions.py`, and `download_data.py` to a new package `raw_feed_download_utils` in order to allow them to be used from other analyses. Exclude functions that are exclusively useful for generating Conveyal network bundles.

* **Question or Goal**: Allow `conveyal_update` functions to be run from other analyses.

* **Data Required**:
  * Currently have: na
  * Need:  na

* **Research Required**: na

* **Metrics**: na

* **Expected Outputs / Findings**: Functions that can be used to get GTFS feeds in a more flexible manner

mrtopsyt:
Considering this a draft for now, per [meeting with Rae](https://gist.github.com/mrtopsyt/cb5134865c040ca2f9bc1b26bdbfe0e9) this may want to be scoped differently.

mrtopsyt:
Would also like to work with @edasmalchi to scope what's necessary here, and how we should prioritize this

mrtopsyt:
Convert existing code to individual functions as defined:
```
# new file get_feeds.py
generate_feed_info(
    target_date: str,
    lookback_period: dt.timedelta | None,
    filter_bbox: BoundingBoxDict | BoundingBox | None,
    filter_geometry: gpd.GeoDataFrame | None,
    filter_geometry_id: str | None,
) -> pd.DataFrame:
    """
    Get a DataFrame containing feeds, filtered by region. Can be passed to `download_feeds` and `download_feeds_region` to download feeds. 

    Params: 
    - target_date: the date to search for feeds on. Should be formatted as a string in YYYY-MM-DD.
    - lookback_period: the time before the period to look for feeds in, if no feed is valid on `target_date`.
    - `filter_geometry`: A GDF containing geometry to spatially filter by. If this is not `None`.
    - `filter_geometry_id`: A column of `filter_geometry_id` to tag rows by. if this is not `None`, the output DF will have an additional column `region`.

    Returns: A DataFrame with the following comments
      - `key`: Unsure
      - `date`: The most recent date within lookback_period on which a valid feed is available Will be equal to `target_date` if `lookback_period` is false
[...]

PR #1629, opened 2025-08-16
Add script for RTA bulk feed generation

mrtopsyt:
Create a script to generate RTA feeds in bulk. Once the `conveyal_update` refactor in #1591 is complete, this will allow for running scripts based on a geometry.

### Changes
- Add script for RTA bulk feed generation
- Removed old proof of concept notebook
- Made `filter_non_rt_trips` a public function, since it is used in the script, and added a basic test
- Minor refactors to reduce unnecessary GCS/warehouse hits


edasmalchi (approved):
looks good!

Issue #1639, opened 2025-08-19
Data Task - Stop-Level GTFS Schedule Summary


KatrinaMKaiser:
## Research Task
Create a quick stop-grain table with columns for peak-hour headway, total trips, total routes for use in a Transit Stop Enhancement Priority pilot tool.

* **Relevant PR** (if applicable):  

* **Question or Goal**: 
From Adrienne St John:
> We have an opportunity to add a use case to one of Caltrans’ two GenAI pilot projects, and we have a transit-related concept.  The current tool being developed is intended to focus on safety of vulnerable road users (VRUs or bikes and peds). Project Delivery is interested in adding on an additional use case to serve our teams, and our preferred concept is to identify and prioritize transit stop locations on the SHS for improvements/amenities. The tool could analyze transit stops based on factors like location in an underserved community, extreme climate, proximity to schools or medical facilities, … and could reference Replica’s data for pedestrian and bicyclist user volumes. It could help provide prioritized lists of locations and, based on any guidance we prepare for it, it could recommend a level or type of transit stop improvement.

This table would contain several of the variables used for prioritization, to measure intensity of stop use.

* **Data Required**:
  * Currently have: previous code for #1099 and #1123 . Use `stop_times` from schedule data. 
  * Need:  harmonize some of the data transformation steps and export one weekday's worth of stop data as a geojson

* **Research Required**:  

* **Metrics**: AM Peak headway, total trips-per-day, total routes that stop at the stop. 
    * Bonus metric if time: use a measure of spatial proximity (short walk distance) to find interagency transfer points, useful for future work. 

* **Expected Outputs / Findings**: stop-grain table with this data. **Due date end-of-August**. Potential foundation for future warehouse data product.

KatrinaMKaiser:
Shweta leading, Amanda supporting/reviewing

shweta487:
https://github.com/cal-itp/data-analyses/pull/1649

PR #1649, opened 2025-08-25
number of trips, routes and peak hour headway per stop analyzed

shweta487:
https://github.com/cal-itp/data-analyses/issues/1639


amandaha8:
I added my comments in a new notebook on the same branch: `ahsc_grant/amanda_Stop_Trip_Route_Analysis.ipynb`

ahsc_grant/amanda_Stop_Trip_Route_Analysis.ipynb
KatrinaMKaiser, line 1: what happens if two agencies both have the same stop id value for two different stops? check your groupby. Also interested in seeing other summary stats on the other variables (frequencies, statistical distribution)

KatrinaMKaiser (changes requested):

PR #1653, opened 2025-08-28
Use Poetry for depenedncy management in realizable_transit_accessibility

mrtopsyt:
`realizable_transit_accessibility` requires multiple internal and external packages beyond what is provided in JupyterHub, therefore creating a need for additional dependency management. This PR shifts to using Poetry to manage these dependencies

- Create a `pyproject.toml` for `realizable_transit_accessbility` including `shared_utils`,  `segment_speed_utils`, their unspecified dependencies, as well as other dependencies
- Create a `poetry.lock` corresponding with the `pyproject.toml`
- Remove the old `requirements.txt` file.


realizable_transit_accessibility/pyproject.toml
```diff
@@ -0,0 +1,40 @@
+[project]
+name = "retrospective_feed_generation"
+version = "0.0.0"
```
raebot, line 3: Since this is not package mode, I think version isn't strictly required.

raebot (approved):
Looks good to me. Installs as expected. One comment in there–you may not need the version property.

realizable_transit_accessibility/pyproject.toml
mrtopsyt: Thanks for catching, I removed that line!

Issue #1660, opened 2025-09-03
[data-analyses] Move archivable files that use siuba to archive repo


raebot:
Complete notes on what to archive or delete or leave be for now are in https://caltrans.sharepoint.com/:w:/r/sites/DOTPMPHQ-DDSContractors/_layouts/15/Doc.aspx?sourcedoc=%7B3F72806B-37F0-4EBE-8ED2-F07387DA39C9%7D&file=archiving-data-analyses.docx&action=default&mobileredirect=true

raebot:
@KatrinaMKaiser should I make the archive repo public? What license should we put on it? I see mostly AGPL-3.0 on other repos.

PR #1669, opened 2025-09-04
Archive and remove old files with siuba

raebot:
Resolves https://github.com/cal-itp/data-analyses/issues/1660
Files moved to new `data-analyses-archive` repo with https://github.com/cal-itp/data-analyses-archive/pull/1

As part of reducing siuba usage and our migration away from using it, we identified files and directories that could be archived. We can continue to access this code for reference, but it doesn't need to be on our radar for siuba migration.


edasmalchi (approved):
awesome, thanks!

Issue #1646, opened 2025-08-21
Refine Use of Explored NTD Variables


KatrinaMKaiser:
## Research Task
We completed #1580 with some variables from the NTD, and have additional questions to resolve before adding new variables.

* **Relevant PR** (if applicable):  #1626 

* **Question or Goal**: 
    * Mode & Service: We previously used these to define the grain of the data. These should be used for classifying/model-fitting. 
        * There are many modes: Can we group modes so there are fewer of them (e.g. fixed-guideway vs not)? Too many dummy variables for a category leads to overfitting/too many clusters. 
        * What's the best way to use Service? Is it as a dummy variable? Is it numeric "proportion of service(VRH or M)" that's directly operated? Explore the impacts of both.
    * ~~If we group by only Agency (flattening, fewer rows), how do we aggregate the other classification variables before we normalize them? Is adding sufficient? Do we need to average any?~~
      * moved to issue #1683 
    * Are there ways to get more interactive/easy-to-read visualizations?
        * If it takes significant time, break this one out
    * ~~Fixed number of clusters vs optimized number: Let's write code to implement the "elbow method" or similar to get the best number of clusters instead of too many/few.~~ 
        * ~~If it takes significant time, break this one out.~~
        * Moved to issue #1647 

* **Data Required**:
  * Currently have: Pre-processed analysis table with a grain of agency-mode-service type. Old EDA notebook. 
  * Need:  

* **Research Required**:  

* **Metrics**: what specific calculations are you looking to find?  

* **Expected Outputs / Findings**: 
    * Table aggregated to agency-grain. 
    * Summary stats for the classification variables after table is flattened and those variables are aggregated/turned into dummies. 
    * Quick look at which agencies are at the top and bottom end of the distributions of those variables. 
    * Samples of which agencies are in specific clusters. 
    * Dendrograms.

PR #1675, opened 2025-09-09
Transit peer group research: explore possible feature columns 

csuyat-dot:
Issue:
- #1646 

Notebook that holds notes from Google Machine Learning Crash Course, specifically about categorical and numerical feature columns. (feature-crosses, log scaling, binning )

Categorized NTD feature columns into different groups. Also explore separating modes by fixed guideway and non-fixed guideway. Explored what "flattening" NTD data looks like after aggregating by:
-  mode/service. each row will be total upt/vrh/vrm by agency per year
- mode/service/year. Each row will be total upt/vrh/vrm for 5 years per agency

Tested 3d scatter plot using plotly express


transit_agency_peer_groups/utils_transit_peer_groups.py
```diff
@@ -0,0 +1,61 @@
...
+)
+    pipeline = Pipeline(
+    [
+        ("preprocessing", preprocessor),
+        ("clustering", AgglomerativeClustering(n_clusters=cluster_num, linkage="ward"))
+    ]
+)
+    data_fit = data.copy() # why do i need to copy/clone?
+
+    data_fit["cluster_name"] = pipeline.fit_predict(data_fit)
```
KatrinaMKaiser, line 34: Does this mean the cluster name is the predicted value?

transit_agency_peer_groups/1646_explore_ntd_variables.ipynb
KatrinaMKaiser, line 1: - Trolleybus is fixed guideway - it has a pantograph attached to overhead wires
- you should simply group the purchased transportation - taxi/TNC with the purchased transportation generally
- It was interesting to see in the multi-year data, Metro is in one cluster pre-pandemic and another post-pandemic. This plus the correlation matrics tells me that we should consolidate to one year of data or aggregate across years, but we definitely shouldn't leave all years loose. To be explored more in #1683 

The last thing I'd like to see on this PR is: 
- Clustering on 2023 data only
- re-categorize the modes into fixed route, non-fixed-route, and other
- re-categorize the service into DO or PT generally
- sum up the numerical columns based on the id cols and the new categorical variables you just created (not the old ones), so you have one row per agency-groupedmode-groupedservice
- try running the clustering, 10 is still fine for now
- print out a sample of 5 rows from each cluster with their characteristics
-  print summary stats grouped by by cluster (like mean UPT for example)

Issue #1616, opened 2025-08-12
[data-analyses] Migrate more complex use of siuba


raebot:
Work with a data analyst on migrating a file with siuba where more that just the super simple functions (`filter`, `collect`) are in use.

PR #1678, opened 2025-09-10
Archive some of sb125_analyses/

raebot:
While de-siubafying `sb125_analyses/quick_mts_jobs/quick_mts_jobs.ipynb`, @edasmalchi and I realized it made more sense just to archive this work and also `sb125_analyses/corridor_study` since it is unlikely to be use again. 

Relates to https://github.com/cal-itp/data-analyses-archive/pull/3
Relates to https://github.com/cal-itp/data-analyses/issues/1616


edasmalchi (approved):
looks good, thanks!

PR #1707, opened 2025-09-25
[shared_utils] Improve catalog lookup in catalog_utils

raebot:
This PR modifies `catalog_utils.get_catalog_file` so that it's able to find a given catalog even if you happen to have your `data-analyses` repo nested within you home directory. For instance `~/workspace/calitp/data-analyses` instead of assuming everyone locates their repos at the top of home like `~/data-analyses`.


ohrite (approved):
🍐w/ @raebot

PR #1766, opened 2025-11-10
Add End of Year Calculations to Holiday Service Mismatch Query

jiaxiliu95:
This PR adds End of Year analysis to the holiday service mismatch query. Same logic as Thanksgiving analysis.

For End of Year analysis, 12/17 (Wednesday) is picked as regular service reference, and the 12/13 weekend is picked as reduced service reference.


vevetron (approved):
Looks good.  Maybe describe the ratios at the top or somewhere. Red, etc.

I did not check calculations this time.

Issue #1770, opened 2025-11-13
Remove siuba from `gtfs_utils_v2`


raebot:
Remove siuba from https://github.com/cal-itp/data-analyses/blob/main/_shared_utils/shared_utils/gtfs_utils_v2.py

PR #1780, opened 2025-11-19
Remove siuba from `shared_utils/gtfs_utils_v2`

raebot:
This PR removes siuba from `shared_utils/gtfs_utils_v2` by converting siuba queries to use SQLAlchemy ORM model querying. It also adds test coverage for the impacted functions, using pytest-recording for requests to BigQuery.

Relates to [#1770](https://github.com/cal-itp/data-analyses/issues/1770)

### DIFF
--- a/_shared_utils/README.md
+++ b/_shared_utils/README.md
@@ -19,13 +19,22 @@ These shared utility functions are quality-of-life improvements for analysts as 
 ## Contributing
 ### Testing
 
-Some code is this directory has test coverage and there's an intention to expand coverage as needed, particularly before and around when changes are made to existing code. These tests are an important step towards providing some regression proofing for our shared code. If changes are made to code that has test coverage, GitHub Actions will fail the build. At that point, you will need to fix the code or update the tests to match new desired behavior.
+Some code is this directory has test coverage and there's an intention to expand coverage as needed, particularly when changes are made to existing code.
+These tests are an important step towards providing some regression proofing for our shared code. GitHub Actions will fail the build and alert us if changes
+are made to code that has test coverage where assertions were not satisfied. At that point, you will need to either fix the code or, if new behavior is desired,
+update the tests to match.
 
-Many of these tests exercise code that requests data from BigQuery datasets. In order to make tests run quickly and not drive up costs, these tests use [pytest-recording](https://pypi.org/project/pytest-recording/) to make recordings (cassettes) of the actual BigQuery requests and their responses. This allows us to reuse existing recordings when we run tests instead of repeatedly having to make a network request.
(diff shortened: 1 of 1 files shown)


_shared_utils/shared_utils/models/bridge_organization_x_headquarters_county_geography.py
```diff
@@ -5,7 +5,7 @@
 Base = declarative_base()
 
 
-class BridgeOrganizationsXHeadquartersCountyGeography(Base):
+class BridgeOrganizationXHeadquartersCountyGeography(Base):
```
raebot: Renaming to match singular model name convention

_shared_utils/shared_utils/gtfs_utils_v2.py
```diff
@@ -481,43 +413,34 @@ def get_stop_times(
             If True, return pd.DataFrame
     """
     check_operator_feeds(operator_feeds)
-
     trip_id_cols = ["feed_key", "trip_id"]
 
     if trip_df is None:
```
raebot: All callers of this function in the repo pass `trip_df`, so this clause isn't strictly needed, but it seemed fine to leave in.

_shared_utils/shared_utils/gtfs_utils_v2.py
```diff
@@ -481,43 +413,34 @@ def get_stop_times(
...
-
-    if not isinstance(trips_on_day, pd.DataFrame):
-        stop_times = (
-            tbls.mart_gtfs.dim_stop_times()
-            >> filter_operator(operator_feeds, include_name=True)
-            >> inner_join(_, trips_on_day, on=trip_id_cols)
-            >> filter_custom_col(custom_filtering)
-            >> subset_cols(stop_time_cols)
-        )
+        trip_df = get_trips(selected_date=selected_date, operator_feeds=operator_feeds, trip_cols=trip_id_cols)
```
raebot: Again, no callers are currently using this. In any case, I thought it would be good to simplify this function by removing `get_df=False` from the `get_trips` call so that it returns a dataframe, matching the `trips_df` argument type and eliminating the need to have special clauses for non-dataframe.

_shared_utils/shared_utils/models/fct_daily_schedule_feed.py
```diff
@@ -5,7 +5,7 @@
 Base = declarative_base()
 
 
-class FctDailyScheduleFeeds(Base):
+class FctDailyScheduleFeed(Base):
```
raebot: Renaming to match singular model name convention

Issue #1777, opened 2025-11-17
november 2025 gtfs_funnel/open_data


edasmalchi:
Run gtfs_funnel and open data publishing workflow for November 2025.

Issue #1608, opened 2025-08-08
Research Task - Publish speedmap segments on open data portal


edasmalchi:
## Research Task

Publish speedmap segments on open data portal

* **Expected Outputs / Findings**: what do you hope to have created when this is complete?

https://gis.data.ca.gov/datasets/4937eeb59fdb4e56ae75e64688c7f2c0_0/explore uses speedmap-style segments (with interpolated segments between far apart stops)

edasmalchi:
https://github.com/cal-itp/data-analyses/commit/b4a1c7fdc984f8a2e7e7aa04d5ff979fa61b6045

KatrinaMKaiser:
Did we say @mrtopsyt could help with this (metadata etc)?

PR #1784, opened 2025-11-20
Nov open data

edasmalchi:
#1777 
#1608

### DIFF
--- a/high_quality_transit_areas/assemble_hqta_points.py
+++ b/high_quality_transit_areas/assemble_hqta_points.py
@@ -31,6 +31,7 @@ from update_vars import (
 )
 
 gcsgp = GCSGeoPandas()
+fs = get_fs()
 catalog = intake.open_catalog("*.yml")
 
 


tiffanychu90:
@edasmalchi: if there's a rule being applied that you want added to the exception, that can be done [here](https://github.com/cal-itp/data-analyses/blob/main/.pre-commit-config.yaml). I didn't catch it in the GH action, I think it's more about removing variables that aren't used, those have to be manually removed before `git add`, `git commit` again, as they won't be automatically deleted by the precommit hook.

edasmalchi:
I manually removed several unused variables, but the errors that remain look like this. The variable assigned on 115 is used on 119, but these sorts of evaluated strings are apparently intentionally unsupported https://github.com/PyCQA/flake8/issues/1315#issuecomment-819046112

> high_quality_transit_areas/assemble_hqta_points.py:115:9: F841 local variable 'schedule_gtfs_dataset_keys' is assigned to but never used

https://github.com/cal-itp/data-analyses/blob/79f21239a28636024d94bbd28d4573eab661b2b1/high_quality_transit_areas/assemble_hqta_points.py#L113-L122

Issue #1822, opened 2025-12-16
Deprecate GTFS Digest preprocessing scripts: `gtfs_funnel / rt_vs_schedule / gtfs_digest`


tiffanychu90:
## Deprecate scripts in various folders
1. `gtfs_funnel`
* aggregate trips to route-direction
* aggregate trips to operator
* standardize and clean route names for time-series 

2.  `rt_vs_schedule`      
* aggregate vp trips to route-direction
* aggregate vp trips to operator 
 
3. `gtfs_digest`   
* `merge_data.py` for concatenating route-direction single dates into 1 file
* `merge_operator_data.py` for concatenating operator summaries for single dates into 1 file
* `merge_operator_service.py` for creating hourly arrivals viz
* remove yamls that check which operators are present - these are part of the broader monitoring and alerting efforts that occur more upstream
   
4. Clean up `Makefiles`
* Remove these from the `Makefile` as they don't need to be run as part of the workflow

## Deprecate datasets `gtfs_analytics_data.yml`
* Comment these out for now
* Surface which other scripts are calling these as inputs, this should error

## Remove files in GCS
* Future TODO: Once data products point at new warehouse tables and are supported for some time, remove these parquets in GCS
* The commented out filepaths will should be programmatically removed with a script.

PR #1823, opened 2025-12-16
Deprecate gtfs funnel to digest preprocessing

tiffanychu90:
* #1822


tiffanychu90:
> Is there a warehouse replacement for `gtfs_funnel/track_publish_dates.py` or can we keep that one around a little longer? I use it to patch in previous dates for [speedmaps](https://github.com/cal-itp/data-analyses/blob/2ee7f1bd1e921e8be6d70832a6b0d183f91fa02a/ca_transit_speed_maps/update_vars_index.py#L77-L79) and [hqta](https://github.com/cal-itp/data-analyses/blob/2ee7f1bd1e921e8be6d70832a6b0d183f91fa02a/high_quality_transit_areas/lookback_wrappers.py#L8)

Oops - let's reinstate this to use in HQTA and speedmaps. Hoping open data will not need these to patch it in!

Issue #1854, opened 2026-01-06
Upgrade to JupyterBook 2.0


raebot:
Investigate if upgrading JupyterBook to 2.0 for portfolio sites resolves some accessibility errors and, unless upgrading would introduce issues, proceed.

Conversation about this here https://teams.microsoft.com/l/message/19:d74c4b926d554caaab9bf36157e01a99@thread.tacv2/1764627778269?tenantId=621b0a64-1740-43cc-8d88-4540d3487556&groupId=4e1aead1-1390-42d6-bb7a-1045529c5fe1&parentMessageId=1764627778269&teamName=DOT%20PMP%20HQ%20-%20DDS%20%2B%20Contractors&channelName=Data%20Office%20Hours&createdTime=1764627778269&ngc=true

<img width="758" height="1105" alt="Image" src="https://github.com/user-attachments/assets/7bf90bde-49f6-471e-a127-05fd537448b6" />

PR #1918, opened 2026-02-05
Upgrade JupyterBook to v2

raebot:
This PR upgrades JupyterBook to v2 and also removes and gitignores compiled portfolio sites. Many of the file changes in this PR are a result of removing the portfolio site files, so there is quite a bit of noise. 

You can see the result of this upgrade on staging for these portfolio sites:
* [AHSC Ridership Prediction Dashboard](https://analysis-staging.dds.dot.ca.gov/ahsc/) - ahsc
* [HQ Transit Corridor & Major Bus Stop Explorer](https://analysis-staging.dds.dot.ca.gov/hqta_explorer/) - hqta
* [SB125 Fund Split Analysis](https://analysis-staging.dds.dot.ca.gov/sb125_fund_split_analysis/) - sba125_fund_split_analysis

When attempting to deploy other sites, I ran into the following errors that analysts will need to address in the notebooks before we can deploy.

* district_digest - error where NB cell has no output
* gtfs_digest - ValueError: Unable to determine data type for the field "Percent of Trips with Vehicle Positions"
* new_transit_metrics - needs to be migrated off of siuba
* ntd_annual_ridership_report - needs to be migrated to us GCSPandas helpers - got the ArrowInvalid error
* ntd_monthly_ridership - error importing label_visualization from shared_utils.portfolio_utils
* rt_trip_updates_operator_metrics - needs to be migrated to us GCSPandas helpers - got the ArrowInvalid error
* rt_trip_updates_stop_metrics - needs to be migrated to us GCSPandas helpers - got the ArrowInvalid error
* rt - error in %%capture_parameters cell in ca_transit_speed_maps/01_new_speedmaps.ipynb
* thruway_intercity_bus - issue with vegafusion[embed] install

Relates to https://github.com/cal-itp/data-analyses/issues/1854

### DIFF
--- a/high_quality_transit_areas/branching_derived_intersections.py
+++ b/high_quality_transit_areas/branching_derived_intersections.py
@@ -9,7 +9,7 @@ from _utils import append_analysis_name
 from calitp_data_analysis.gcs_geopandas import GCSGeoPandas
 from calitp_data_analysis.gcs_pandas import GCSPandas
 from calitp_data_analysis.geography_utils import CA_NAD83Albers_m
-from IPython.display import Markdown, display
+from IPython.display import HTML, display
 from segment_speed_utils import gtfs_schedule_wrangling, helpers
 from tqdm import tqdm
 from update_vars import (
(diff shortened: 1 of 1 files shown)


edasmalchi:
> MD H3 header not rendering correctly on HQTA branching page
This is a known issue with JBv2 and they care close to enabling it. I filed https://github.com/jupyter-book/jupyter-book/issues/2585#issuecomment-3881061082 and was given this other issue to track progress https://github.com/jupyter-book/mystmd/issues/1026. @edasmalchi what do you think about waiting for their solution? Otherwise you will need to find another way to generate those visualization headers for now.

for my use case, I don't mind the raw markdown while they fix upstream. Wonder if dynamically outputting some HTML rather than Markdown would render properly?

PR #1944, opened 2026-02-26
RT trip updates - Jan 2026 stop (+route filter) report

tiffanychu90:
### Major changes for GTFS RT MSA stop report

Each report changed substantially to figure out how to get a lot of stop-level metrics into visualizations that made sense. 
* route filters - stops are more easily interpreted with a route in mind
* averages for metrics - use all Jan 2026 weekdays for this report. This will be resilient to how we choose to process RT data / frequency to create these dbt models moving forward.
* Explore more interactive options across `altair, geopandas, great_tables`
   * `altair v6` makes interactive parameters more easily inherited across charts
   * demo this with a route dropdown + controlling multiple charts
   * 1 map with multiple layers for easier understanding of stops. 
   * use `great_tables` nanoplot + `great_table_extras` to do even more fun tabular stats
* Closes https://github.com/cal-itp/data-infra/issues/4101
* Past stop reports: #1732, #1680

### thoughts around better maps / charts
* the `gdf.explore` for multiple layers works well, and can even map multiple routes. but without a clear all / select all option, it can be clunky to use for large operators. 
* `altair.mark_geoshape` does work with stops, but right now, the interactivity is limited. Current state is the route dropdown works to create a static map of stops, but it's hard to see stop location even without zoomable scroll. Desired interactivity would be to use the route dropdown and create a mini map that zooms into the relevant stops and is scrollable. [altair_tiles](https://vega.github.io/altair_tiles/intro.html) is doing more in this space!


tiffanychu90:
Even though there's `altair` dependency management error, there's a lot of work in the space around JupyterBook v2 and interactive charts, so we might be upgrading lots of packages anyway

PR #1968, opened 2026-03-18
Migrate to uv workspace with local dependencies

jparr:
Changes to a uv managed project (use kernel: `Pyproject Local`)
ref: cal-itp/data-infra#1882 
Internalizes as many of the dependencies as possible into this repo: 

- calitp-data-analysis is now internal to this repo but a separate package to preserve imports (import calitp_data_analysis.package still works)
- move packages that were installed into jupyterhub image internal to repo - should reduce image churn and need for testing
- uses `pyproject-local-kernel` to automatically switch venvs when you switch project. unlocks multiple projects and repos in the same workspace

depends on `Prototype Image - 2026.3.18, Python 3.11` 

To try it out 

1. launch with `Prototype Image - 2026.3.18, Python 3.11`
2. switch to this branch (`cd data-analyses && git checkout mov/1882-internalize-dependencies`)
3. run `uv sync` (should take a couple minutes but will be fast in the future)
4. use the kernel: `Pyproject Local`
5. try stuff in the repo 

If you run into a problem let me know. it may be expected because a followup to this work is to bring all the individual projects into the uv workspace at which point they can install their own dependencies or we may plan to extract them into their own repos.


erikamov:
I was able to build the Portfolio Index on Staging using the new image and `uv` on this branch.

Once I started the Prototype image, I ran these commands:
1. `uv sync`
2. `make add_precommit`
3. `uv sync --group portfolio`
4. `glogin`
5. `uv run python portfolio/portfolio.py index --deploy`

I added `uv run` to the python command, so we probably need to update the commands on Readmes and make files.

mrtopsyt:
Hi! @csuyat-dot and I pair-ed to run Python scripts locally on Windows, and we worked on top of this branch. We were able to make a non-Portfolio script run successfully, but there were a few issues. Since they mostly related to Windows-specific issues, I'm not sure if they're in the scope of this issue, or if we should create new issues for these.

- `_shared_utils.tzset` is invoked in the body of`shared_utils.rt_utils` but is not supported on Windows. This was previously documented in a different context in https://github.com/cal-itp/data-analyses/issues/2029, so might not need to be handled in this issue. The quick fix is to comment out the reference to it, but I'm unsure of whether that causes unexpected behavior.
- We noted one potentially missing dependency (`openpyxl`, needed for `pd.read_excel` in some scripts).
- quarto-cli is required by `_shared_utils`, but when installing it on Windows UV gives an **extremely** long error. The start and end are below. I'm actually unsure what quarto-cli is used for, but when it was removed from the pyproject.toml for shared_utils, we were able to install it properly.

``` 
Resolved 264 packages in 26ms
   Building quarto-cli==1.6.40
  ├ù Failed to build `quarto-cli==1.6.40`
  Γö£ΓöÇΓû╢ The build backend returned an error
  Γò░ΓöÇΓû╢ Call to `setuptools.build_meta.build_wheel` failed (exit code: 1)

      [stdout]
      Current working directory:
      C:\Users\s163107\AppData\Local\uv\cache\sdists-v9\pypi\quarto-cli\1.6.40\Pv8jr6cOJCIRROi8hrYXn\src
      running bdist_wheel
      running build
      running build_py
      Downloading and installing quarto-cli binaries...
      Downloading
      https://github.com/quarto-dev/quarto-cli/releases/download/v1.6.40/quarto-1.6.40-win.zip
      creating build\lib\quarto_cli
... (20,000 lines!)
      hint: This usually indicates a problem with the package or the build environment.
  help: `quarto-cli` (v1.6.40) was included because `cal-itp-data-analyses:dev` (v0.1.0) depends on
        `shared-utils` (v5.0.0) which depends on `quarto-cli`

mrtopsyt:
I was also successful using the Pyproject Local image to run Notebooks from the [tsp-analysis](github.com/cal-itp/tsp-analysis) repo in JH. I converted a pyproject.toml intended for Poetry to UV using `uvx migrate-to-poetry`, and then was able to run the Notebook by running `uv sync` then selecting pyproject Local as the kernel. This is a really nice functionality that will make JH easier to use for me!

erikamov:
> * openpyxl

I think those could be new issues. Thanks for testing.

Issue #1876, opened 2026-01-14
Develop plan for splitting portfolio sites into their own repo(s)


raebot:
We want to split portfolio sites away from the data-analyses repo because they're public-facing, whereas a lot of other work in the repo is not public-facing. Also they are some of the largest files in the repo, slowing down git operations in the repo as whole.

Outcomes
* propose a couple plans for how to split: thematic grouping or one repo for each site?
* description of deployment process: 
    * staying the same as how sites are deployed today? 
    * do we want to setup automated deployment flows in github actions?
* do we want to keep including built portfolio sites in git history in the same way?

jparr:
Considering two pathways: 

1. separate portofio as a separate repo
2. separate each individual portfolio as a separate repo

Need a picture of how each looks and the trade-offs.

note: consensus preference seems to be the second.

jparr:
Notes from conversation with @edasmalchi is to explore whether we could adapt a site definition (portfolio/sites/rt.yml)
to point to a repository:

```yaml
repository: https://github.com/cal-itp/my-portofolio-site
directory: ./ca_transit_speed_maps/
notebook: ./ca_transit_speed_maps/01_new_speedmaps.ipynb
```

PR #2059, opened 2026-05-28
Cal-itp portfolio build/deploy is a standalone CLI

jparr:
Converted the portfolio build/deploy process into a standalone CLI as a workspace package. This means that it can be run from inside any portfolio project regardless of the location.

It is 'batteries included' and incorporates the magic `%%capture_parameters%%` as well as the builder functions in `shared_utils/portfolio_utils`

ref: #1876


jparr:
needs README and cleanup as well as the accessibility check.
see: https://github.com/cal-itp/data-analyses/tree/mov/test-calitp-portfolio for the changes I made to individual portfolios in order to test this.

PR #2174, opened 2026-08-18
Compare VP- and TU-based stop arrivals data

### LINKED ISSUE #2165: Research Task - Compare VP and Stop Times Data
## Research Task

As an analyst, I am interested in having a single source of truth I can use for the times when transit vehicles arrive at a stop. However, this data is available from both the `gtfs_funnel` `rt_schedule_stop_times` table (VP based) and the `fct_stop_times_metrics` table (TU based). Create a Notepad that generates the following results for a sample route on the same day on each of SBMTD, Samtrans, and San Diego MTS.

- What days are both datasets available for
- Mean / median / standard deviation of the difference between the two tables
- $$R^2$$ value of the correlation between the two

* **Question or Goal**: Determine a single dataset for use for TSP performance metrics

* **Data Required**:
  * Currently have: Both datasets
  * Need: May need to generate new data if they are not both available on the same day

mrtopsyt:
### Changes

- Added notebook and sql queries to compare stop arrivals generated from Vehicle Positions (VP) in the speeds pipeline and from Trip Updates (TU) in the `fct_stop_times_metrics` table on dates where both are available.

### Summary of Findings
- TU-based data is generally available for a greater number of trips than VP-based data
- For different agencies checked (SamTrans, SFMTA, SBMTD, SacRT), 83-90% of stop arrivals available for both agencies matched within 60 sec. Lower levels of alignment were anecdotally associated with MTC Regional Feeds.
- For SamTrans specifically, TU stop arrivals snapped to within an integer multiple of 60 seconds from a schedule, which was not observed for VP.
- For all agencies checked, TU data was associated with a 5-10x greater incidence of non-monotonic arrivals, however the magnitude of these non-monotonic cases was smaller. Note that the VP data is subject to data cleaning steps designed to decrease non-monotonicity, whereas the TU-based data is not


mrtopsyt:
@hhmckay @edasmalchi, this is the notebook I presented to you! Merging because it's all new files, but wanted to ping you in case you want to take a look

